PR #1, opened 2016-07-12
Merge presentation

philandstuff:
This merges the openregister/presentation repo into this repository.

In order for this to work, I needed to do the following:

 - I made the functional tests:
   - start the dropwizard app server themselves, using
     DropwizardAppRule, rather than relying on gradle to start the app
     server for you
   - use a custom dns resolver rather than hacking the Host header
     manually
   - mint data properly, via the mint API, rather than messing around
     with the database
 - I changed the presentation DAO classes to refer to the tables that
   already exist -- eg current_entry_number not total_entries

There is also some outstanding work that still needs to be done after
this commit:

 - I had to `@Ignore` some tests for the moment. we should get them
   running again.
 - I also have disabled the conformance tests (they are commented
   out). We should re-enable them
 - there is a substantial amount of duplication (eg: multiple
   TestEntryDAO objects) that should be addressed
 - the package structure is a mess and should be addressed
 - the Cardinality.fromId method can possibly be removed?

but, overall, it works!


chr1sdi:
👍 works. Outstanding work mentioned in the comment will be done on a different branch

PR #4, opened 2016-07-14
Remove register configuration stuff from RequestContext

philandstuff:
This removes the register configuration methods from RequestContext.

When we configured the register by sniffing the Host: header, it made
some kind of sense for this code to live here.  But now we get it from
yaml configuration instead, let's rip it all out.


philandstuff:
there are similar methods on ThymeleafView that could be removed but I didn't have time to do them.

chr1sdi:
👍

PR #9, opened 2016-07-26
X forwarded proto

philandstuff:
This supersedes openregister/presentation#255 by doing the Right Thing instead.

(warning: "scheme" and "protocol" are used interchangably here).

Currently, the app uses the scheme in the http request to the app server
to determine what scheme to use in URLs rendered as links (in html and
ttl formats, among others).

This is problematic because the scheme used to access an individual app
server might differ from the scheme used by a user to access the edge
server.

This can be resolved by use of the X-Forwarded-Proto header, which is
set by (among other things) Amazon Elastic Load Balancers (ELBs).  This
header is used to report what the edge request protocol was, rather than
using the individual app server protocol.


chr1sdi:
👍

PR #10, opened 2016-07-26
reinstate conformance tests

philandstuff:
This commit re-enables the conformance tests that were removed during
the early stages of merging the apps.  In doing this, I've copied over
the previous strategy of starting up an app in the gradle file and
cloning/running the conformance tests.  However I've done some changes:

1) where possible, I define tasks and dependencies between tasks, rather
than as doFirst() or doLast() blocks.  In particular, the conformance
tests are run as part of the build by making the `check` task depend on
the conformance task. [1]

This has the advantage, among other things, that you are able to run the
conformance tests separately by running `./gradlew runConformanceTests`
now.

2) I have created a separate config file for running the app in
conformance test mode -- `conformance-app-config.yaml`.  This was
necessary because the conformance test data is based on schools and so
needs to set `register: school` in the config file.  I took the
opportunity to run it on a different port to avoid clashes with the app
servers run via DropwizardAppRules in the main tests.

3) The conformance tests actually found a bug!  The /registers.yaml
resource wasn't working, because the RegisterDetailView has changed to
no longer extend AttributionView as it used to.  It doesn't make sense
to make RegisterDetailView extend io.dropwizard.views.View because it
never gets rendered as HTML.  So instead I have changed YamlWriter to be
able to render any kind of object, not merely subclasses of View.

[1]: http://blog.proxerd.pl/article/gradle-s-check-task

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,15 +70,73 @@ assemble {
     }
 }
 
+task startAppForConformance << {
+    startApp()
+}
+
+void startApp() {
+    ProcessBuilder builder = new ProcessBuilder(
+            "java",
+            "-cp",
+            sourceSets.test.runtimeClasspath.join(":"),
+            "uk.gov.register.RegisterApplication",
+            "server",
+            "src/test/resources/conformance-app-config.yaml"
+    ).directory(projectDir.absoluteFile)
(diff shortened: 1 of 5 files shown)


chr1sdi:
👍

PR #12, opened 2016-07-26
bump to dropwizard 1.0.0

philandstuff:
Relatively painless upgrade.  Release notes:
http://www.dropwizard.io/1.0.0/docs/about/release-notes.html#v1-0-0-jul-26-2016

The main difference that affected us is that the dropwizard-java8 stuff
has been merged into the main dropwizard code, so some packages were
renamed and some stuff went away (such as the Java8Bundle, which is
included by default).


chr1sdi:
👍

PR #26, opened 2016-08-03
Display Register proof coming from verifiable log in fingerprint on the home page

chr1sdi:
https://trello.com/c/EB1530E1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ before_script:
 - psql -c 'create database test_openregister_java;' -U postgres
 - psql -c 'create database ft_openregister_java;' -U postgres
 after_failure:
-- cat /home/travis/build/openregister/openregister-java/stderr.txt
-- cat /home/travis/build/openregister/openregister-java/stdout.txt
+- cat $TRAVIS_BUILD_DIR/stderr.txt
+- cat $TRAVIS_BUILD_DIR/stdout.txt
 deploy:
 - provider: s3
   access_key_id: AKIAINSAHPOLGXFFYGMQ
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,9 @@ before_script:
 - psql --version
 - psql -c 'create database test_openregister_java;' -U postgres
 - psql -c 'create database ft_openregister_java;' -U postgres
+after_failure:
+- cat /home/travis/build/openregister/openregister-java/stderr.txt
+- cat /home/travis/build/openregister/openregister-java/stdout.txt
 deploy:
 - provider: s3
   access_key_id: AKIAINSAHPOLGXFFYGMQ
--- a/src/main/java/uk/gov/register/db/EntryMerkleLeafStore.java
+++ b/src/main/java/uk/gov/register/db/EntryMerkleLeafStore.java
@@ -8,18 +8,20 @@ import uk.gov.register.core.Entry;
 import uk.gov.verifiablelog.store.MerkleLeafStore;
 
 public class EntryMerkleLeafStore implements MerkleLeafStore {
-
     private final EntryQueryDAO entryDAO;
     private final EntryIteratorDAO entryIteratorDAO;
+    private final boolean hasEntries;
+
 
     public EntryMerkleLeafStore(EntryQueryDAO entryDAO) {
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -30,6 +30,9 @@ before_script:
 - psql --version
 - psql -c 'create database test_openregister_java;' -U postgres
 - psql -c 'create database ft_openregister_java;' -U postgres
+after_failure:
+- cat /home/travis/build/openregister/openregister-java/stderr.txt
```
philandstuff: @chr1sdi tells me that we'd like to use TRAVIS_BUILD_DIR here

src/main/java/uk/gov/register/db/EntryMerkleLeafStore.java
```diff
@@ -8,18 +8,20 @@
...
     public EntryMerkleLeafStore(EntryQueryDAO entryDAO) {
         this.entryDAO = entryDAO;
         this.entryIteratorDAO = new EntryIteratorDAO(entryDAO);
+        this.hasEntries = totalLeaves() > 0;
     }
 
     @Override
     public byte[] getLeafValue(int i) {
-        return bytesFromEntry(entryIteratorDAO.findByEntryNumber(i + 1));
+        return hasEntries ? bytesFromEntry(entryIteratorDAO.findByEntryNumber(i + 1)) : new byte[]{0x00};
```
philandstuff: why is `getLeafValue(i)` ever called when `hasEntries` is true?

src/main/java/uk/gov/register/resources/HomePageResource.java
```diff
@@ -1,34 +1,60 @@
 package uk.gov.register.resources;
 
 import io.dropwizard.views.View;
+import jdk.nashorn.internal.runtime.options.Option;
```
philandstuff: did you mean to import this?

src/main/java/uk/gov/register/resources/HomePageResource.java
```diff
@@ -1,34 +1,60 @@
...
     @Produces({ExtraMediaType.TEXT_HTML})
-    public View home() {
-        return viewFactory.homePageView(recordDAO.getTotalRecords(), entryDAO.getTotalEntries(), entryDAO.getLastUpdatedTime());
+    public View home() throws NoSuchAlgorithmException {
+        Optional<RegisterProof> registerProof = Optional.empty();
+        try {
+            entryDAO.begin();
+            VerifiableLog verifiableLog = createVerifiableLog(entryDAO);
+            String rootHash = bytesToString(verifiableLog.currentRoot());
+            registerProof = Optional.of(new RegisterProof(rootHash));
```
philandstuff: under what circumstances can `registerProof` legitimately be `empty()`? my understanding is that it can always be computed, even for a register of 0 entries.

src/main/java/uk/gov/register/resources/HomePageResource.java
```diff
@@ -1,34 +1,60 @@
...
+        } finally {
+            entryDAO.rollback();
+        }
+
+        return viewFactory.homePageView(
+                recordDAO.getTotalRecords(),
+                entryDAO.getTotalEntries(),
+                entryDAO.getLastUpdatedTime(),
```
philandstuff: I wonder if `lastUpdatedTime` should be `Optional`? An empty register doesn't have a last updated time but is still a legitimate register. 💭

src/main/java/uk/gov/register/views/HomePageView.java
```diff
@@ -69,4 +76,13 @@ public String getLinkToRegisterRegister() {
                 getRegisterId()
         );
     }
+
+    public String getRegisterJson() throws JsonProcessingException {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
```
philandstuff: does it make sense to (at some point) add this ObjectMapper to the hk2 DI container?

.travis.yml
chr1sdi: worked like a charm :)

PR #28, opened 2016-08-03
Allow openregister-java to be deployed to heroku

philandstuff:
I got this deployed to heroku. This was a wonderful yak shave:

 - heroku is based on logging to console rather than to file
 - trying to enable a logger by configuring it with `type: console` resulted in weird errors
 - these errors were caused by our fat jar not concatenating metadata properly
 - rather than fixing our home-grown fat jar code, I replaced it with the shadowJar plugin.

After this PR, there is a thin jar at `build/libs/openregister-java.jar` and a far jar at `build/libs/openregister-java-all.jar`.  The fat jar is used for everything -- in particular, starting the app for conformance test, and for copying into the .zip file which is uploaded to S3 for CodeDeploy.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: java -jar build/libs/openregister-java-all.jar server heroku.yaml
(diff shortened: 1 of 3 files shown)


chr1sdi:
👍

PR #32, opened 2016-08-08
Entry timestamp assignment

chr1sdi:
https://trello.com/c/D6OwBNRo/331-app-should-assign-timestamps-not-database

This includes: 
  - timestamp should be assigned from the app
  - merge Entry classes (mint and register)
  - change timestype type to int
  - fix tests
  - db migration scripts


src/main/java/uk/gov/register/core/FatEntry.java
```diff
@@ -1,6 +1,6 @@
 package uk.gov.register.core;
 
-import uk.gov.mint.Entry;
+import uk.gov.register.core.Entry;
```
michaelabenyohai: I don't think this import is required.

michaelabenyohai:
The "sha:256:" prefix for item-hash is missing from some views/representations:

# /entries
# /entry/1.ttl

I believe Entry.getSha256hexFull needs to be used instead of Entry.getSha256hex in the below files:

# entries.html
# EntryTurtleWriter.java

I also notice that from a previous commit (in merging) the "item-resource" field has changed to "item-resources - please can you change this back to "item-resource" in EntryTurtleWriter.java.

michaelabenyohai:
Suggest that we also change getSha256hex() to getSha256hexFull in Item class so that the method names are consistent between Item and Entry (and doesn't confuse us). But would also mean updating getSha256hex to getSha256hexFull for Item in all views/representations.

michaelabenyohai:
I'm happy with this. As well as the above comments I have also tested that loading works as expected.

We should chat with @philandstuff and agree strategy for deployment since the database needs to be updated first.

PR #34, opened 2016-08-11
Remove dead code

philandstuff:
Since we moved to Dropwizard 1.0, Java 8 is supported by default.  This
means that the Jdk8Module and the JavaTimeModule are automatically
supported by Jackson.newObjectMapper(), and we don't need to add them
separately.


chr1sdi:
👍

PR #35, opened 2016-08-11
remove dynamic register proof from home page

philandstuff:
For large registers, each time we redeploy we'll trash the in-memory
cache of merkle tree nodes.  If we have the root hash on the homepage,
then each time we redeploy, we won't be able to render the homepage
until the in-memory cache has been rebuilt.

Reverting back to the static mockup for the time being.


chr1sdi:
👍  and 😢

PR #42, opened 2016-08-31
fix AssetsBundle 404s, remove RegisterNameExtractor

philandstuff:
Any 404 within the AssetsBundle path (eg /assets/not-found ) will
currently break.  This is because the `renderedCopyrightText` variable
is not set, causing the template engine to break.

Also, the error handler was the last place still getting the register
name from the Host header rather than from the register configuration.
I took the opportunity to replace that.


chr1sdi:
👍  goodie

PR #43, opened 2016-08-31
don't show PASSED tests in build output

philandstuff:
When a test fails, it's hard to see which one failed because it's lost
in the noise of all the PASSED tests in the output.

We originally added this output to give us confidence that the tests
were actually being run.  I don't think we need that any more.


chr1sdi:
👍  kapow

PR #48, opened 2016-09-23
Spike: introduce register abstraction

philandstuff:
**DO NOT MERGE!** This PR, in its current form, introduces *painful* performance penalties.

This is an attempt to introduce interfaces `Register` and `RegisterReadOnly` that capture in one place the kinds of operations and queries you can do on a register. There is an implementation, PostgresRegister, which captures all of the existing behaviour.

The PostgresRegister is super inefficient because it goes back to INSERTing rows one-by-one rather than getting proper benefit from JDBC batch inserts. This could be improved by having the PostgresRegister batch things up internally in memory and periodically send stuff off in a batch insert operation.

I think this leads to interesting questions about where database transactions get managed (in this object? in the consumer?).

Thoughts?


philandstuff:
I pushed a couple of updates that (are supposed to) move transaction handling in to the PostgresRegister object. I'm not sure whether it actually does what I want -- when I log out the SQL stuff, it seems to be doing some nested transaction committing stuff.

philandstuff:
I've pushed some more commits that resolve the performance and transaction handling issues.

Any comments? Anyone? :tumbleweed:

philandstuff:
rebased.

michaelabenyohai:
The VerifiableLogService appears to have its own EntryQueryDAO independent of the one used by the RegisterDAO. Could we get rid of EntryMerkleLeafStore and have RegisterDAO implement MerkleLeafStore? The EntryIteratorDAO can be use the EntryQueryDAO attached to the handle in RegisterDAO, similar to how it's already done for DestinationDBUpdateDAO. Then we can pass the RegisterDAO to the VerifiableLogService and remove bind(entryQueryDAO) from RegisterApplication.

philandstuff:
> Is it correct that Register 'is a' ReadOnlyRegister? Could Register be a separate interface - maybe WritableRegister?

I don't think it's incorrect in the sense of [Liskov substitutability](https://en.wikipedia.org/wiki/Liskov_substitution_principle). The intention is that a `RegisterReadOnly` is simply a read-only view on to a `Register`, much like `Collections.unmodifiableList()` offers a read-only view of a `List`.  Maybe this could be communicated better?

We could instead have independent "read" and "write" interfaces, but I think we'll soon get to a point where we'll need to consume both in one class. In particular, in the external representation spike, we want to write a bunch of items and entries and then read a root hash to assert that we've got the correct value.

philandstuff:
> Does PostgresRegister have any specific link to Postgres? Should the Register type in the model just show its relationships to its component items etc?

@andyjpb made this point too. I think you're both right.

>RegisterDao only interacts with the database through other xxxDao objects. Is it more of a service than a Dao?

Yes, probably. I wonder if RegisterDao will maybe go away after further refactoring?

> Generally as transactions encompass >1 call to a Dao, should all GetHandle implementing classes live in a service layer?

Yes, I think so.

I think this shows that introducing this interface shines light on further refactoring that is desirable 😄

philandstuff:
> The VerifiableLogService appears to have its own EntryQueryDAO independent of the one used by the RegisterDAO. Could we get rid of EntryMerkleLeafStore and have RegisterDAO implement MerkleLeafStore? The EntryIteratorDAO can use the EntryQueryDAO attached to the handle in RegisterDAO, similar to how it's already done for DestinationDBUpdateDAO. Then we can pass the RegisterDAO to the VerifiableLogService and remove bind(entryQueryDAO) from RegisterApplication.

Yes, the VerifiableLogService was just chucked in wholesale without further refactoring at this point. I didn't want to do this in this PR as it's already quite a big changeset and I haven't made things any worse.  But you rightly raise further opportunities for refactoring 😀

PR #51, opened 2016-09-29
Link to register history from front page

michaelabenyohai:
Configure the link, which may not be present for all registers.

### DIFF
--- a/src/main/java/uk/gov/register/RegisterConfiguration.java
+++ b/src/main/java/uk/gov/register/RegisterConfiguration.java
@@ -6,6 +6,7 @@ import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.db.DataSourceFactory;
 import uk.gov.organisation.client.GovukClientConfiguration;
 import uk.gov.register.auth.RegisterAuthenticatorFactory;
+import uk.gov.register.configuration.RegisterContentPagesConfiguration;
 import uk.gov.register.configuration.RegisterDomainConfiguration;
 import uk.gov.register.configuration.RegisterNameConfiguration;
 import uk.gov.register.configuration.ResourceConfiguration;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/register/RegisterConfiguration.java
```diff
@@ -57,6 +57,11 @@ public String getRegisterDomain() {
     @JsonProperty
     private boolean enableDownloadResource = false;
 
+    @SuppressWarnings("unused")
+    @Valid
+    @JsonProperty
+    private String historyPageUrl;
```
philandstuff: Is it not possible to make this an `Optional<String>`?

src/main/java/uk/gov/register/configuration/RegisterContentPages.java
```diff
@@ -0,0 +1,36 @@
...
+        return "https://www.gov.uk/guidance/using-registers-to-build-a-service";
+    }
+
+    @SuppressWarnings("unused, used from template")
+    public String getSpecificationUrl() {
+        return "http://openregister.github.io/specification/";
+    }
+
+    @SuppressWarnings("unused, used from template")
+    public String getRegisterHistoryPageUrl() {
```
philandstuff: Can we just provide the `Optional` direct to the template and have the template call `isPresent` and `get`? That's a pattern we do elsewhere - see `copyright.html` for an example

src/main/java/uk/gov/register/views/ViewFactory.java
```diff
@@ -24,24 +26,24 @@
...
     private final RegisterDomainConfiguration registerDomainConfiguration;
+    private final RegisterContentPages registerContentPages;
 
     @Inject
     public ViewFactory(RequestContext requestContext,
                        ItemConverter itemConverter,
                        PublicBodiesConfiguration publicBodiesConfiguration,
                        GovukOrganisationClient organisationClient,
-                       RegisterDomainConfiguration domainConfiguration,
+                       RegisterConfiguration registerConfiguration,
```
philandstuff: Rather than injecting the whole `RegisterConfiguration` class, I think it's better to inject a minimal interface (in this case, we might want to create a `RegisterContentPagesConfiguration` interface?).  These smaller interfaces are much easier to stub out in tests and it makes it clearer what the class is actually depending on.

michaelabenyohai:
@philandstuff I've pushed some commits to address your comments

PR #53, opened 2016-10-03
Introduce more domain concepts

philandstuff:
Having introduced the register abstraction in #48, this continues that work by introducing more abstractions for pieces of our domain:

 - EntryLog, which encapsulates the append-only list of entries and the verifiable log of proofs
 - ItemStore, which encapsulates the store of items addressed by hash
 - RecordIndex, which encapsulates the linking of primary keys to current entries

They are still somewhat interrelated through transactions: you shouldn't be able to add a dangling entry or item so we often need to pass the same Handle into the ItemStore and the EntryLog to ensure they are both updated together, atomically.

As @johnollier suggested on #48, the name "PostgresRegister" is increasingly inaccurate at this stage but I can't think of a better name.


src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -20,107 +20,113 @@
...
     @Override
     public List<Record> max100RecordsFacetedByKeyValue(String key, String value) {
-        return registerDAO.findMax100RecordsByKeyValue(key, value);
+        return dbi.withHandle(h -> recordIndex.findMax100RecordsByKeyValue(h, key, value));
     }
 
     @Override
     public RegisterProof getRegisterProof() throws NoSuchAlgorithmException {
-        return verifiableLogService.getRegisterProof();
+        return dbi.withHandle(entryLog::getRegisterProof);
```
karlbaker02: We'll want to double check that this is wrapped in a transaction, or explicitly open a transaction

src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -20,107 +20,113 @@
...
     }
 
     @Override
     public EntryProof getEntryProof(int entryNumber, int totalEntries) {
-        return verifiableLogService.getEntryProof(entryNumber, totalEntries);
+        return dbi.withHandle(handle ->
```
karlbaker02: We'll want to double check that this is wrapped in a transaction, or explicitly open a transaction

src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -20,107 +20,113 @@
...
+                entryLog.getEntryProof(handle, entryNumber, totalEntries));
     }
 
     @Override
     public ConsistencyProof getConsistencyProof(int totalEntries1, int totalEntries2) {
-        return verifiableLogService.getConsistencyProof(totalEntries1, totalEntries2);
+        return dbi.withHandle(handle ->
```
karlbaker02: We'll want to double check that this is wrapped in a transaction, or explicitly open a transaction

philandstuff:
rebased.

src/main/java/uk/gov/register/RegisterApplication.java
```diff
@@ -101,15 +95,17 @@ public void run(RegisterConfiguration configuration, Environment environment) th
             @Override
             protected void configure() {
                 bind(entryQueryDAO).to(EntryQueryDAO.class);
```
michaelabenyohai: Can we remove this line now?

philandstuff: yes! opened #54 for this.

PR #55, opened 2016-10-18
Specify register for conformance tests in build.gradle

michaelabenyohai:
And therefore ensure the register name does not need to be slurped from the URL


karlbaker02:
👍

PR #57, opened 2016-10-27
Add analytics

chr1sdi:
https://trello.com/c/seTp90X4/421-add-support-for-google-analytics

### DIFF
--- a/src/main/resources/assets/js/analytics-download-type.js
+++ b/src/main/resources/assets/js/analytics-download-type.js
@@ -1,9 +1,11 @@
-var elems = document.getElementsByClassName("js-download");
+(function(){
+    var elems = document.getElementsByClassName("js-download");
 
-setupEventAnalytics(elems, "Data", "download", function(e){
-    return e.target.getAttribute("data-download-type") || "unrecognised";
-});
+    setupEventAnalytics(elems, "Data", "download", function(e){
+        return e.target.getAttribute("data-download-type") || "unrecognised";
+    });
 
-setupEventAnalytics(elems, "Data", "download", function(e){
-    return "data-download";
-});
+    setupEventAnalytics(elems, "Data", "download", function(e){
+        return "data-download";
+    });
+})();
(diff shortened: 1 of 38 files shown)


src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
```diff
@@ -0,0 +1,121 @@
...
+        Boolean docIncludesAnalyticsId = trackingIdElem != null;
+        Boolean docIncludesMainAnalytics = doc.getElementById("analytics-main") != null;
+        Boolean docIncludesExtLinksAnalytics = doc.getElementById("analytics-external-links") != null;
+
+        assertThat(docIncludesAnalyticsId, equalTo(shouldIncludeAnalytics));
+        assertThat(docIncludesMainAnalytics, equalTo(shouldIncludeAnalytics));
+        assertThat(docIncludesExtLinksAnalytics, equalTo(shouldIncludeAnalytics));
+
+        if (shouldIncludeAnalytics) {
+            assertThat(docIncludesAnalyticsId, equalTo(true));
```
karlbaker02: This is a duplicate of line 82: if `shouldIncludeAnalytics` is `true` before then we're going to run the same check again here

src/main/resources/assets/js/analytics-download-type.js
```diff
@@ -0,0 +1,11 @@
+(function(){
+    var elems = document.getElementsByClassName("js-download");
+
+    setupEventAnalytics(elems, "Data", "download", function(e){
+        return e.target.getAttribute("data-download-type") || "unrecognised";
+    });
+
+    setupEventAnalytics(elems, "Data", "download", function(e){
```
karlbaker02: Is this second function intended? Same signature, same name.

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
chr1sdi: was added for clarity of intention, will remove it if the test is clear without it

src/main/resources/assets/js/analytics-download-type.js
chr1sdi: it's an invocation of the setup for 2nd event on data download click

PR #58, opened 2016-10-28
Memoization rollback

karlbaker02:
Adding ability to rollback MemoizationStore if a transaction fails (for example, assert-root-hash fails).


src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
...
+import uk.gov.verifiablelog.store.memoization.MemoizationStore;
+import java.util.HashMap;
+
+public class TransactionalMemoizationStoreImpl implements TransactionalMemoizationStore {
+    private final MemoizationStore instance;
+    private final HashMap<Integer, HashMap<Integer, byte[]>> temporaryEntries;
+    private Integer currentEntryCount;
+
+    public TransactionalMemoizationStoreImpl(
+            MemoizationStore instance) {
```
chr1sdi: need to break the line here?

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+import static org.hamcrest.core.Is.is;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
+
+public class TransactionMemoizationStoreImplTest {
```
chr1sdi: SUT is `TransactionalMemoizationStoreImpl` maybe we can rename the filename

src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
```diff
@@ -31,6 +29,9 @@ private PostgresDriverTransactional(Handle handle, MemoizationStore memoizationS
         this.stagedEntries = new ArrayList<>();
         this.stagedItems = new LinkedHashSet<>();
         this.stagedCurrentKeys = new HashMap<>();
+
+        // Initialise the entry cutoff now that have have our PostgresDriver created
+        ((TransactionalMemoizationStore)memoizationStore).setCurrentEntryCount(this.getTotalEntries());
```
chr1sdi: we can pass in `TransactionalMemoizationStore` in private constructor and remove this casting, what do you think?

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
...
+        this.currentEntryCount = currentEntryCount;
+    }
+
+    @Override
+    public void put(Integer start, Integer size, byte[] value) {
+        if ((start + size) <= currentEntryCount) {
+            return;
+        }
+
+        if (!temporaryEntries.containsKey(start)) {
```
chr1sdi: `temporaryEntries` does this mean stagedHashes?

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
chr1sdi: can we rename `instance` to something else pls it's not very descriptive, suggests it's a self instance and not the decorated one.

src/main/java/uk/gov/register/core/TransactionalMemoizationStore.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.register.core;
+
+import uk.gov.verifiablelog.store.memoization.MemoizationStore;
+
+public interface TransactionalMemoizationStore extends MemoizationStore {
+    void commitEntries();
+    void rollbackEntries();
+    void setCurrentEntryCount(int entryCutoff);
```
chr1sdi: - different param name `currentEntryCount` in the impl
  - any reason the field's type is `Integer` but the param `int`?

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public void setCurrentEntryCount(int currentEntryCount) {
+        this.currentEntryCount = currentEntryCount;
+    }
+
+    @Override
+    public void put(Integer start, Integer size, byte[] value) {
+        if ((start + size) <= currentEntryCount) {
```
chr1sdi: would be more readable if this condition was assigned to a nicely named bool just like https://github.com/openregister/openregister-java/pull/58/files#diff-cb527ab39d508d10a6982c973dcdc212R38

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
...
+            return;
```
chr1sdi: shouldn't this pass those to the underlying store just like in `get`?
```if ((start + size) <= currentEntryCount) {
    instance.put(start, size, value);
    return;
}```
somink like that, what do you think?

File: src/main/java/uk/gov/register/core/TransactionalMemoizationStore.java [hunk 8]
```diff
@@ -0,0 +1,9 @@
+package uk.gov.register.core;
```
chr1sdi: can this live in `store` package? don't know, what do you think?

src/main/java/uk/gov/register/core/TransactionalMemoizationStore.java
```diff
@@ -0,0 +1,9 @@
...
+
+import uk.gov.verifiablelog.store.memoization.MemoizationStore;
+
+public interface TransactionalMemoizationStore extends MemoizationStore {
```
chr1sdi: does this need to be an interface? is there any change volatility around this that we are going to have multiple implementations of it?

src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
```diff
@@ -46,11 +47,27 @@ protected PostgresDriverTransactional(Handle handle, MemoizationStore memoizatio
...
+            handle.begin();
+
+            PostgresDriverTransactional postgresDriver = new PostgresDriverTransactional(handle, transactionalMemoizationStore);
             callback.accept(postgresDriver);
             postgresDriver.writeStagedData();
-        });
+
+            handle.commit();
+            transactionalMemoizationStore.commitEntries();
+        } catch (Exception ex) {
```
chr1sdi: you will make this `Exception ex` sad if you make it disappear silently :(

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import uk.gov.verifiablelog.store.memoization.MemoizationStore;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
```
chr1sdi: quite generous imports, should we import only things we use? any thoughts?

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
+package uk.gov.register.core;
+
+import uk.gov.verifiablelog.store.memoization.MemoizationStore;
+import java.util.HashMap;
+
+public class TransactionalMemoizationStoreImpl implements TransactionalMemoizationStore {
+    private final MemoizationStore instance;
+    private final HashMap<Integer, HashMap<Integer, byte[]>> temporaryEntries;
+    private Integer currentEntryCount;
```
chr1sdi: implicitly 0, shouldn't this explicitly in constructor?

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+            HashMap<Integer, byte[]> hashesBySize = temporaryEntries.get(start);
+            return hashesBySize.get(size);
+        }
+
+        return instance.get(start, size);
+    }
+
+    @Override
+    public void commitEntries() {
```
chr1sdi: 'entries' is a bit ambiguous name, maybe we can rename this method

src/main/java/uk/gov/register/core/TransactionalMemoizationStoreImpl.java
```diff
@@ -0,0 +1,68 @@
...
+                instance.put(start, size, hash);
+            });
+        });
+        temporaryEntries.clear();
+
+        setCurrentEntryCount(temporaryEntries.size() + currentEntryCount);
+    }
+
+    @Override
+    public void rollbackEntries() {
```
chr1sdi: same as `commitEntries`

src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
```diff
@@ -46,11 +47,27 @@ protected PostgresDriverTransactional(Handle handle, MemoizationStore memoizatio
...
+            handle.setTransactionIsolation(TransactionIsolationLevel.SERIALIZABLE);
+            handle.begin();
+
+            PostgresDriverTransactional postgresDriver = new PostgresDriverTransactional(handle, transactionalMemoizationStore);
             callback.accept(postgresDriver);
             postgresDriver.writeStagedData();
-        });
+
+            handle.commit();
+            transactionalMemoizationStore.commitEntries();
```
chr1sdi: what is going to happen in the exception block with `handle.rollback()` if `transactionalMemoizationStore.commitEntries();` chucks an exception?
maybe the ordering of committing can solve this

src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
```diff
@@ -31,6 +29,9 @@ private PostgresDriverTransactional(Handle handle, MemoizationStore memoizationS
         this.stagedEntries = new ArrayList<>();
         this.stagedItems = new LinkedHashSet<>();
         this.stagedCurrentKeys = new HashMap<>();
+
+        // Initialise the entry cutoff now that have have our PostgresDriver created
```
chr1sdi: a bit confusing comment, entry cutoff is slightly ambiguous

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
+
+public class TransactionMemoizationStoreImplTest {
+    public TransactionalMemoizationStoreImpl createTransactionMemoizationStore(
```
chr1sdi: line break quite confusing

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+        store.setCurrentEntryCount(entryCount);
+
+        return store;
+    }
+
+    @Test
+    public void getShouldNotReturnHashWhenNoHashExistsForGivenStartIndexAndTreeSize() {
+        byte[] hash = "22905e52ce72f3880bddd5564966b15b017a038d84de41c390b2aafe28fd8452".getBytes();
+        MemoizationStore internalStore = mock(MemoizationStore.class);
+        when(internalStore.get(0, 2)).thenReturn(hash);
```
chr1sdi: is this needed?

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+        store.put(0, 1, hash);
+        byte[] result = store.get(0, 1);
+
+        verify(internalStore, never()).get(0, 1);
+        assertThat(Arrays.equals(result, hash), is(true));
+    }
+
+    @Test
+    public void putShouldNotAddHashWhenDoingSoWouldRewriteHistory() {
```
chr1sdi: is this a requirement?

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+        }).when(internalStore).put(anyInt(), anyInt(), captor.capture());
+        TransactionalMemoizationStoreImpl store = createTransactionMemoizationStore(internalStore, 0);
+
+        store.put(0, 1, hash);
+
+        assertThat(addedHashes, hasSize(0));
+    }
+
+    @Test
+    public void commitEntriesShouldCommitAllHashesToUnderlyingMemoizationStore() {
```
chr1sdi: `entries` is a bit ambiguous

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Test
+    public void putShouldNotPersistHashDirectlyToMemoizationStore() {
+        ArrayList<byte[]> addedHashes = new ArrayList<>();
+        byte[] hash = "22905e52ce72f3880bddd5564966b15b017a038d84de41c390b2aafe28fd8452".getBytes();
+        MemoizationStore internalStore = mock(MemoizationStore.class);
+        when(internalStore.get(0, 1)).thenReturn(hash);
+
+        ArgumentCaptor<byte[]> captor = ArgumentCaptor.forClass(byte[].class);
```
chr1sdi: seems like it can be extracted to a nicer sounding method as it's used in a lot of places, what do you think?

src/test/java/uk/gov/register/core/TransactionMemoizationStoreImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+        store.rollbackEntries();
+
+        // Commit any and all entries to prove that temporary entries have been cleared
+        store.commitEntries();
+
+        assertThat(addedHashes, hasSize(0));
+    }
+
+    @Test
+    public void setCurrentEntryCountShouldUpdateCurrentEntryCount() {
```
chr1sdi: this test does not test that, there would be no difference in test if I set `store.setCurrentEntryCount(5)` yet it makes difference in SUT

chr1sdi:
last thing, can you rebase on top of latest master please :)
cheers

PR #59, opened 2016-10-31
Fixing faceted search to return 404 if supplied field name does not e…

karlbaker02:
…xist for the given register


chr1sdi:
tested and can be rebased, can you rebase before we merge please?

michaelabenyohai:
Would it be possible to move `registerContainsField()` into `PostgresRegister` (we should rename this to something without Postgres in it). This is because it doesn't require any Postgres operation and therefore I feel like it does not need to be all the way down in `PostgresDriver`. 

It would also mean that we pass `RegistersConfiguration` to `PostgresRegister` instead of to `PostgresDriver`. This makes sense since we already pass `RegisterNameConfiguration` to `PostgresRegister`.

Please could you also create an interface such as `RegisterFieldsConfiguration` and pass only the required field data in - this makes it easier to Mock and test the `PostgresRegister` object.

src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -1,7 +1,11 @@
 package uk.gov.register.core;
 
+import com.google.common.collect.Lists;
+import uk.gov.register.RegisterConfiguration;
```
michaelabenyohai: Please can you clean up the imports - this one isn't needed.

src/main/java/uk/gov/register/store/postgres/PostgresDriver.java
```diff
@@ -7,22 +7,21 @@
...
-import java.util.Collection;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 
 public abstract class PostgresDriver implements BackingStoreDriver {
 
     protected final MemoizationStore memoizationStore;
+    private final HashMap<String, ArrayList<String>> registerFieldLookup;
+
```
michaelabenyohai: This is no longer used. There should be no changes to this file as part of this PR.

src/main/java/uk/gov/register/store/postgres/PostgresDriver.java
```diff
@@ -114,7 +114,7 @@ public int getTotalRecords() {
     }
 
     @Override
-    public List<Record> findMax100RecordsByKeyValue(String key, String value) {
+    public List<Record> findMax100RecordsByKeyValue(String registerName, String key, String value) {
```
michaelabenyohai: registerName is not used in this method

src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
```diff
@@ -7,6 +7,7 @@
 import org.skife.jdbi.v2.exceptions.CallbackFailedException;
 import org.skife.jdbi.v2.tweak.HandleCallback;
 import org.skife.jdbi.v2.tweak.HandleConsumer;
+import uk.gov.register.configuration.RegistersConfiguration;
```
michaelabenyohai: Tidy up imports, this isn't used.

src/test/java/uk/gov/register/functional/store/postgres/PostgresDriverTransactionalFunctionalTest.java
```diff
@@ -17,6 +18,7 @@
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.core.Is.is;
+import static org.mockito.Mockito.mock;
```
michaelabenyohai: Tidy up imports

src/test/java/uk/gov/register/store/postgres/PostgresDriverTransactionalTest.java
```diff
@@ -3,10 +3,10 @@
...
 import org.mockito.ArgumentCaptor;
-import uk.gov.register.core.Entry;
-import uk.gov.register.core.Item;
-import uk.gov.register.core.Record;
+import uk.gov.register.configuration.RegistersConfiguration;
+import uk.gov.register.core.*;
 import uk.gov.register.db.*;
+import uk.gov.register.exceptions.NoSuchFieldException;
 import uk.gov.register.views.RegisterProof;
 import uk.gov.verifiablelog.VerifiableLog;
```
michaelabenyohai: Tidy up imports. There should be no change to this file.

src/test/java/uk/gov/register/core/PostgresRegisterTest.java
```diff
@@ -0,0 +1,43 @@
...
+import uk.gov.register.exceptions.NoSuchFieldException;
+import uk.gov.register.store.BackingStoreDriver;
+import uk.gov.register.store.postgres.PostgresDriverTestBase;
+
+import java.util.ArrayList;
+
+import static org.mockito.Mockito.*;
+
+public class PostgresRegisterTest extends PostgresDriverTestBase {
+    private RegisterNameConfiguration registerNameConfiguration;
```
michaelabenyohai: This does not need to extend PostgresDriverTestBase as these tests do not need the mocked functionality in that class.

karlbaker02: Good spot, thanks.

PR #61, opened 2016-11-02
Memoization rollback with previous PR feedback incorporated

karlbaker02:


src/main/java/uk/gov/register/core/TransactionalMemoizationStore.java
```diff
@@ -0,0 +1,49 @@
...
+        HashMap<Integer, byte[]> hashesBySize = stagedHashes.get(start);
+        hashesBySize.put(size, value);
+    }
+
+    @Override
+    public byte[] get(Integer start, Integer size) {
+        byte[] hash = null;
+
+        if (stagedHashes.containsKey(start)) {
+            hash = stagedHashes.get(start).get(size);
```
philandstuff: why not just `return hash` here?

karlbaker02: This is what I had meant to do, thanks for pointing out.

philandstuff: in which case, I'd find `if (hash != null) { return hash; }` less confusing here than the ternary currently used below

src/main/java/uk/gov/register/core/TransactionalMemoizationStore.java
```diff
@@ -14,30 +15,21 @@ public TransactionalMemoizationStore(MemoizationStore memoizationStore) {
...
+        stagedHashes.put(Pair.of(start, size), value);
     }
 
     @Override
     public byte[] get(Integer start, Integer size) {
-        byte[] hash = null;
-
-        if (stagedHashes.containsKey(start)) {
-            hash = stagedHashes.get(start).get(size);
+        if (stagedHashes.containsKey(Pair.of(start, size)))  {
```
chr1sdi: maybe assigning `Pair.of(start, size)` to an variable would be clearer and we can reuse it in the `get`?
somink like: `Pair<Integer, Integer> startSizePair = Pair.of(start, size);`  ? what do you think?

chr1sdi: if we assume that there is not going to be a `null` hash value for any star, size pair we can even compress it to a 1 liner, something like this:
```
public byte[] get(Integer start, Integer size) {
        return stagedHashes.getOrDefault(Pair.of(start, size), memoizationStore.get(start, size));
    }
```

chr1sdi:
👍

PR #63, opened 2016-11-08
Check for dangling entries in register object

michaelabenyohai:
Create exception mapper for NoSuchItemForEntryException and add tests to ensure it is thrown when we attempt to insert and entry that has no corresponding item in the register.


src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -1,9 +1,10 @@
 package uk.gov.register.core;
 
 import com.google.common.collect.Lists;
+import org.thymeleaf.templatemode.ITemplateModeHandler;
```
karlbaker02: This import isn't used.

michaelabenyohai: Fixed, thanks!

src/test/java/uk/gov/register/core/PostgresRegisterTest.java
```diff
@@ -41,4 +45,60 @@ public void findMax100RecordsByKeyValueShouldReturnValueWhenKeyExists() {
...
+
+        PostgresRegister register = new PostgresRegister(registerData, backingStoreDriver, itemValidator);
+
+        try {
+            register.appendEntry(entryNotDangling);
+            register.appendEntry(entryDangling);
+        } catch (NoSuchItemForEntryException ex) {
+        }
+
+        verify(backingStoreDriver, times(1)).insertEntry(any());
```
karlbaker02: Perhaps not strictly necessary, but we're just asserting here that `insertEntry` is called only once, not whether a dangling entry is being inserted.

I'd probably split this out instead as:
```verify(backingStoreDriver, times(1)).insertEntry(entryNotDangling);
verify(backingStoreDriver, never()).insertEntry(entryDangling);```

File: src/test/java/uk/gov/register/core/PostgresRegisterTest.java [hunk 3]
```diff
@@ -41,4 +45,60 @@ public void findMax100RecordsByKeyValueShouldReturnValueWhenKeyExists() {
...
+
+        PostgresRegister register = new PostgresRegister(registerData, backingStoreDriver, itemValidator);
+
+        try {
+            register.appendEntry(entryNotDangling);
+            register.appendEntry(entryDangling);
+        } catch (NoSuchItemForEntryException ex) {
+        }
+
+        verify(backingStoreDriver, times(1)).insertRecord(any(), eq("country"));
```
karlbaker02: Similarly, we're not checking here the record that is inserted is the record we expect to be inserted.

```ArgumentCaptor<Record> record = ArgumentCaptor.forClass(Record.class);
        verify(backingStoreDriver, times(1)).insertRecord(record.capture(), eq("country"));
        assertThat(record.getValue().entry, equalTo(entryNotDangling));```

PR #64, opened 2016-11-09
Check canonicalization of items in itemstore

karlbaker02:
This is an implementation of validating items that are being added to a register to ensure that they are canonicalized. It is based on the following assumptions:

- We will be storing the raw item JSON as part of the `Item`, enabling us to use the output from the  `JsonMapper` and `CanonicalJsonMapper` to determine if an item is in canonical form.
- `Item` should keep the property `content` of type `JsonNode`, as this node is determined when the `Item` is created based on whether the input is in canonical form or not.
- `DataUpload` has been extended to include a new endpoint for loading data without applying canonicalization to `Item`s. Whilst not technically part of this story, it shows how `Item`s will be constructed.
- All `Item`s added to the `ItemStore` will need to be in canonical form, as the `ItemValidator` will throw an exception otherwise. This is the intended behaviour: items added to the register via the 'legacy' endpoint will be canonicalized on their way in, whereas items added via the new endpoint have the requirement to be in canonical form to begin with.
- Downloading data from the register in ZIP format will canonicalize items.

### DIFF
--- a/src/test/java/uk/gov/register/service/ItemValidatorTest.java
+++ b/src/test/java/uk/gov/register/service/ItemValidatorTest.java
@@ -5,9 +5,10 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Test;
 import uk.gov.register.configuration.FieldsConfiguration;
 import uk.gov.register.configuration.RegistersConfiguration;
+import uk.gov.register.core.Item;
 import uk.gov.register.exceptions.ItemValidationException;
-import uk.gov.register.service.ItemValidator;
 import uk.gov.register.util.CanonicalJsonMapper;
+import uk.gov.register.util.JsonMapper;
 
 import java.io.IOException;
 import java.util.Optional;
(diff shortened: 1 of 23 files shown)


src/test/java/uk/gov/register/service/ItemValidatorTest.java
```diff
@@ -102,9 +112,42 @@ public void validateItem_throwsValidationException_givenFieldWithCardinalityMany
...
+        try {
+            itemValidator.validateItem("register", item);
+            fail("Entry is not canonicalized but threw no exception");
+        } catch (ItemValidationException e) {
+            assertThat(e.getMessage(), equalTo("Entry is not canonicalized"));
+        }
+    }
+
+    @Test
+    public void validateItemIsCanonicalized_throwsNoValidationException_whenItemIsCanonicalized() throws IOException {
```
karlbaker02: This test should technically be covered by the below test (`noErrorWhenEntryIsValid`) and can be removed.

PR #65, opened 2016-11-09
Check items for canonicalization when loading register data

karlbaker02:
Now checking that items are in canonical form when loading data into the register


src/test/java/uk/gov/register/util/CanonicalJsonValidatorTest.java
```diff
@@ -0,0 +1,36 @@
...
+    public void validateItemIsCanonicalized_throwsNoValidationException_whenItemIsCanonicalized() throws IOException {
+        String jsonString = "{\"register\":\"aregister\",\"text\":\"some text\"}";
+
+        try {
+            canonicalJsonValidator.validateItemStringIsCanonicalized(jsonString);
+        } catch (SerializationFormatValidationException e) {
+            fail("Entry is canonicalized but exception was thrown");
+        }
+    }
+}
```
michaelabenyohai: There should be a test that checks that an item with white space in it throws an exception (an item with white space is not canonical).

src/main/java/uk/gov/register/util/CanonicalJsonValidator.java
```diff
@@ -0,0 +1,26 @@
...
+        this.jsonMapper = new JsonMapper();
+        this.canonicalJsonMapper = new CanonicalJsonMapper();
+    }
+
+    public void validateItemStringIsCanonicalized(String jsonItem) {
+        byte[] jsonItemBytes = jsonItem.getBytes(StandardCharsets.UTF_8);
+        byte[] nonCanonicalizedBytes = jsonMapper.writeToBytes(jsonMapper.readFromBytes(jsonItemBytes));
+        byte[] canonicalizedBytes = canonicalJsonMapper.writeToBytes(canonicalJsonMapper.readFromBytes(jsonItemBytes));
+
+        if (!Arrays.equals(nonCanonicalizedBytes, canonicalizedBytes)) {
```
michaelabenyohai: We should be comparing jsonItemBytes to canonicalizedBytes.

nonCanonicalizedBytes in this case will have removed any white space that appeared in jsonItemBytes and therefore prevents us from detecting white space in jsonItemBytes. Therefore nonCanonicalizedBytes should be removed and not used.

michaelabenyohai:
I don't think we should have any changes to `DataUpload.java` or to `ObjectReconstructor.java`. Creating a new endpoint that loads canonical data is part of another story and it will be loading RSF lines, not JSON lines. This extra endpoint is therefore not necessary and only be deleted later. The new CanonicalJsonValidator can be used by the person writing the RSF load endpoint.

src/main/java/uk/gov/register/util/CanonicalJsonValidator.java
michaelabenyohai: The definition of canonicalization is here http://openregister.github.io/specification/#sha-256-item-hash, where it talks about not allowing white space.

src/main/java/uk/gov/register/util/JsonMapper.java
```diff
@@ -0,0 +1,33 @@
...
+        try {
+            // Method from http://stackoverflow.com/questions/18952571/jackson-jsonnode-to-string-with-sorted-keys
+            Object obj = objectMapper.treeToValue(jsonNode, Object.class);
+            // for some reason, writeValueAsString(obj).getBytes() doesn't re-escape unicode, but writeValueAsBytes does
+            return objectMapper.writeValueAsBytes(obj);
+        } catch (IOException e) {
+            return ExceptionUtils.rethrow(e);
+        }
+    }
+}
```
michaelabenyohai: Can we remove this class as it's no longer required - CanonicalJsonMapper does not require it as a base class.

michaelabenyohai:
Please can you rebase? This cannot be merged as is.

PR #66, opened 2016-11-10
Checking field exists in SearchResource

karlbaker02:
SearchResource shouldn't redirect to /records if the given field name doesn't exist in the current register, but should instead return a 404.

Refactored the checking of an existence of a field in a register into RegisterFieldsConfiguration, which is used in both SearchResource and PostgresRegister (although the latter may not technically need to be checking  this any longer if all requests first go through SearchResource).

### DIFF
--- a/src/main/java/uk/gov/register/service/RegisterService.java
+++ b/src/main/java/uk/gov/register/service/RegisterService.java
@@ -1,6 +1,7 @@
 package uk.gov.register.service;
 
 import org.skife.jdbi.v2.DBI;
+import uk.gov.register.configuration.RegisterFieldsConfiguration;
 import uk.gov.register.core.PostgresRegister;
 import uk.gov.register.core.Register;
 import uk.gov.register.core.RegisterData;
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/register/resources/SearchResourceTest.java
```diff
@@ -33,11 +43,16 @@ public HttpServletResponse getHttpServletResponse() {
...
-        resource = new SearchResource(requestContext, () -> "school");
+        RegisterMetadata registerMetadata = mock(RegisterMetadata.class);
+        registerData = mock(RegisterData.class);
+        when(registerData.getRegister()).thenReturn(registerMetadata);
+        registerFieldsConfiguration = mock(RegisterFieldsConfiguration.class);
     }
 
     @Test
     public void findSupportsTurtleHtmlAndJson() throws Exception {
+        resource = new SearchResource(requestContext, () -> "school", registerFieldsConfiguration);
```
michaelabenyohai: `resource` isn't used by this test so can be removed here

src/main/java/uk/gov/register/service/RegisterService.java
```diff
@@ -15,18 +16,20 @@
...
+    private final RegisterFieldsConfiguration registerFieldsConfiguration;
 
     @Inject
-    public RegisterService(RegisterData registerData, DBI dbi, MemoizationStore memoizationStore, ItemValidator itemValidator) {
+    public RegisterService(RegisterData registerData, DBI dbi, MemoizationStore memoizationStore, ItemValidator itemValidator, RegisterFieldsConfiguration registerFieldsConfiguration) {
         this.registerData = registerData;
         this.dbi = dbi;
         this.memoizationStore = memoizationStore;
         this.itemValidator = itemValidator;
+        this.registerFieldsConfiguration = registerFieldsConfiguration;
```
michaelabenyohai: You could `new RegisterFieldsConfiguration(registerData)` here and then you wouldn't need to also pass it in the constructor.

karlbaker02: That's true, as we already have access to `registerData` in here. Should registration not live in `RegisterApplication` though?

michaelabenyohai: Good question - I'm trying to avoid the number of registrations in `RegisterApplication` that don't need to be there. However you need it to be there for injection to `SearchResource` so I'm happy to leave it.

PR #71, opened 2016-11-14
Check items are in canonical format when loading from RSF

karlbaker02:


src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -30,7 +39,10 @@ public RegisterCommand newCommand(String s){
             case "add-item":
                 if (parts.length == 2) {
                     try {
-                        Item item = new Item(new ObjectMapper().readTree(parts[1]));
+                        String jsonContent = parts[1];
+                        canonicalJsonValidator.validateItemStringIsCanonicalized(jsonContent);
+                        String itemHash = DigestUtils.sha256Hex(jsonContent.getBytes());
```
michaelabenyohai: Can we use `getBytes(StandardCharsets.UTF_8)`? This would be consistent with the what currently happens in `ObjectReconstructor` for the original /load endpoint.

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -30,7 +39,10 @@ public RegisterCommand newCommand(String s){
...
+                        Item item = new Item(itemHash, new ObjectMapper().readTree(jsonContent));
```
michaelabenyohai: Can we add another method to `ObjectReconstructor`. This will just make sure that we are using consistent methods for the existing /load endpoint and the new one (only difference is we're not using `CanonicalJsonMapper` for the new one).

```
public JsonNode reconstruct(String[] jsonObjectsAsStrings) {
        return new ObjectMapper().readValue(e.getBytes(StandardCharsets.UTF_8), JsonNode.class));
    }
```

src/test/java/uk/gov/register/resources/RegisterCommandReaderTest.java
```diff
@@ -1,34 +1,32 @@
...
 
-import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
 
 @RunWith(MockitoJUnitRunner.class)
 public class RegisterCommandReaderTest {
-
     Class<RegisterCommandList> type = RegisterCommandList.class;
+    private RegisterCommandReader parser;
```
michaelabenyohai: Should this be renamed to `registerCommandReader` or something like `rsfParser`? Only because it looks a bit confusing in the constructor to have a variable called `commandParser` and a variable called `parser`.

src/test/java/uk/gov/register/resources/RegisterCommandWriterTest.java
```diff
@@ -93,7 +97,10 @@ public void writeTo_shouldAddContentDispositionHeader() throws IOException {
         }).when(httpHeadersMock).add(eq("Content-Disposition"), anyObject());
 
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-        RegisterCommandWriter sutCommandWriter = new RegisterCommandWriter();
+        CanonicalJsonMapper canonicalJsonMapper = new CanonicalJsonMapper();
+        CanonicalJsonValidator canonicalJsonValidator = new CanonicalJsonValidator();
+        CommandParser commandParser = new CommandParser(canonicalJsonMapper, canonicalJsonValidator);
+        RegisterCommandWriter sutCommandWriter = new RegisterCommandWriter(commandParser);
```
michaelabenyohai: It might be nice to extract the above lines into a method called `getRegisterCommandWriter()` and extract the code so that we don't duplicate so much.

PR #73, opened 2016-11-16
Check for orphan items during parsing.

johnollier:


src/main/java/uk/gov/register/resources/RegisterCommandReader.java
```diff
@@ -47,7 +40,9 @@ public RegisterSerialisationFormat readFrom(Class<RegisterSerialisationFormat> t
 
     private RegisterSerialisationFormat parseCommands(InputStream commandStream) {
         BufferedReader buffer = new BufferedReader(new InputStreamReader(commandStream));
-        Iterator<RegisterCommand> commands = buffer.lines().map(s -> commandParser.newCommand(s)).iterator();
+        final CommandParser parser = new CommandParser();
+        buffer.lines().forEach(s -> parser.addCommand(s));
```
karlbaker02: Can we keep the previous implementation of creating Iterator<RegisterCommand> ?

johnollier: CommandParser.getCommands does return an iterator. However its true that we are not streaming the input - I'm not sure if we were really doing it before either. Anyway the basic problem is that as we require add item commands to come before append entry commands, the orphan item check has to happen after all the append entry commands have been loaded.

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -45,40 +60,75 @@ public RegisterCommand newCommand(String s){
...
+            } else if (proofs.containsKey(i)) {
+                return new AssertRootHashCommand(proofs.get(i));
+            } else {
+                throw new RuntimeException("No command found for position " + String.valueOf(i));
+            }
+        }).iterator();
+    }
+
+    private void validateOrphanItems() {
+        Set<Item> orphanItems = items.entrySet().stream().filter(kv -> {
```
karlbaker02: Would it be possible just to use a `HashMap` or similar that stores item hashes against number of times that item hash has been referenced by `append-entry` commands? Then you can keep track of what is and isn't being used, and also enforces the required ordering.

src/test/java/uk/gov/register/serialization/CommandParserTest.java
```diff
@@ -26,66 +28,75 @@
...
+    public void shouldFailForOrphanItem() throws Exception {
+        commandParser.addCommand("add-item\t{\"register\":\"ft_openregister_test\",\"text\":\"SomeText\"}");
+        commandParser.addCommand("add-item\t{\"register\":\"ft_openregister_test\",\"text\":\"orphan\"}");
+        commandParser.addCommand("append-entry\t2016-11-02T14:45:54Z\tsha-256:3cee6dfc567f2157208edc4a0ef9c1b417302bad69ee06b3e96f80988b37f254");
+        commandParser.getCommands();
+    }
+
+    @Test(expected = OrphanItemException.class)
+    public void shouldFailForItemAfterEntry() throws Exception {
+        commandParser.addCommand("append-entry\t2016-11-02T14:45:54Z\tsha-256:3cee6dfc567f2157208edc4a0ef9c1b417302bad69ee06b3e96f80988b37f254");
```
karlbaker02: This test is actually testing whether there is a dangling item, so perhaps we should rename it? We should also add a test something along the lines of the following (if this is a valid use case):

```
@Test(expected = OrphanItemException.class)
    public void shouldAddAlreadyExistingItemAsNewEntry() {
        try {
                commandParser.addCommand("append-entry\t2016-11-02T14:45:54Z\tsha-256:3cee6dfc567f2157208edc4a0ef9c1b417302bad69ee06b3e96f80988b37f254");
        commandParser.getCommands();
        } catch (Exception e) {
                fail("Should have added new entry for item previously in register");
        }
    }
```

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -45,40 +60,75 @@ public RegisterCommand newCommand(String s){
...
+        validateOrphanItems();
+
+        return IntStream.range(0, position).mapToObj(i -> {
+            if (entries.containsKey(i)) {
+                return new AppendEntryCommand(entries.get(i));
+            } else if (items.containsKey(i)) {
+                return new AddItemCommand(items.get(i));
+            } else if (proofs.containsKey(i)) {
+                return new AssertRootHashCommand(proofs.get(i));
+            } else {
```
karlbaker02: This `else` statement could be removed, as could the `else if` statements just be `if` statements (minor detail)

src/main/java/uk/gov/register/serialization/CommandParser.java
johnollier: good idea

src/test/java/uk/gov/register/serialization/CommandParserTest.java
johnollier: I didn't think 'dangling' is very clear, so I introduced the term 'orphan' item for an item which no entry refers to. There are actually 2 parts to the requirement viz the item should be referred to by an entry and that entry must come later in the RSF.

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -45,40 +59,84 @@ public RegisterCommand newCommand(String s){
...
+        if (itemHashToEntryCount.containsKey(sha256hex)) {
+            Integer count = itemHashToEntryCount.get(sha256hex);
+            itemHashToEntryCount.replace(sha256hex, count + 1);
+        }
+    }
+
+    public Iterator<RegisterCommand> getCommands() {
+        validateOrphanItems();
+
+        return IntStream.range(0, position).mapToObj(i -> {
```
karlbaker02: If we later find out that `AssertRootHash` is only going to appear at the start and end of an RSF file, we could revert to adding the set of items to the register first, after which we then go through the (ordered) list and add entries. Happy to merge this for the time being though.

PR #74, opened 2016-11-16
Add assert-root-hash statements to RSF when downloading part of a reg…

karlbaker02:
…ister

### DIFF
--- a/external-dependencies.gradle
+++ b/external-dependencies.gradle
@@ -3,7 +3,7 @@ ext {
     jersey_version = '2.23.2'
     jackson_version = '2.8.3'
 
-    verifiableLog = 'verifiable-log:verifiable-log:0.2.0000'
+    verifiableLog = 'verifiable-log:verifiable-log:0.2.77'
 
     dropwizard = [
             "io.dropwizard:dropwizard-core:${dropwizard_version}",
--- a/external-dependencies.gradle
+++ b/external-dependencies.gradle
@@ -3,7 +3,7 @@ ext {
     jersey_version = '2.23.2'
     jackson_version = '2.8.3'
 
-    verifiableLog = 'verifiable-log:verifiable-log:0.1.67'
+    verifiableLog = 'verifiable-log:verifiable-log:0.2.0000'
 
     dropwizard = [
             "io.dropwizard:dropwizard-core:${dropwizard_version}",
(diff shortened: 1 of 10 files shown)


external-dependencies.gradle
```diff
@@ -3,7 +3,7 @@ ext {
     jersey_version = '2.23.2'
     jackson_version = '2.8.3'
 
-    verifiableLog = 'verifiable-log:verifiable-log:0.1.67'
+    verifiableLog = 'verifiable-log:verifiable-log:0.2.0000'
```
karlbaker02: Aware that this version may change but including in this PR for completeness

chr1sdi: So maybe we can put `verifiableLog = 'verifiable-log:verifiable-log:0.2.+'` over here to take the latest. VL does not change that often really and usually if it does then we do rebuild and bump the number over here. They are closely coupled when it comes to triggering changes so it may be ok to put +

src/test/java/uk/gov/register/functional/DataDownloadFunctionalTest.java
```diff
@@ -144,6 +151,17 @@ public void downloadPartialRSF_shouldReturnReturn404ForRsfBoundariesOutOfBound()
         assertThat(response.getStatus(), equalTo(400));
     }
 
+    @Test
+    public void downloadPartialRSF_shouldReturnSameRootHashAsDownloadRSF() {
```
chr1sdi: that's a really cool test, can we drop just hashes being tested so:
`assertThat(fullRsfLines, equalTo(partialRsfLines));` and rename it `shouldReturenTheSameRSF bla bla bla`?

src/test/java/uk/gov/register/service/RegisterSerialisationFormatServiceTest.java
```diff
@@ -54,6 +52,7 @@
     private RegisterCommand appendEntryCommand2;
     private RegisterCommand assertRootEmptyRegister;
 
+
```
chr1sdi: why???? ;)

src/test/java/uk/gov/register/service/RegisterSerialisationFormatServiceTest.java
```diff
@@ -70,6 +69,7 @@ public void setUp() throws Exception {
         appendEntryCommand2 = new AppendEntryCommand(entry2);
         assertRootEmptyRegister = new AssertRootHashCommand(emptyRegisterProof);
 
+
```
chr1sdi: why???? ;)

chr1sdi:
-  that 1 line which can be more readable (if not then it's fine)
 -  don't forget to rebase/update please

chr1sdi:
👍

PR #75, opened 2016-11-18
Primary key in entry

chr1sdi:
part of: https://trello.com/c/9fhVSQL4/415-add-primary-key-into-the-entry-in-openregister-java

note: Conformance tests will fail if https://github.com/openregister/conformance-test/pull/17 not merged


philandstuff:
If I run:

    dropdb openregister_java
    createdb openregister_java
    ./run-application.sh

I get a SQL error: `ERROR: column "key" of relation "entry" already exists`

philandstuff:
the problem is in `create_schema.sh` and was introduced in 398532580e762e3565. I will be refactoring these scripts to fix these issues in #85.

PR #79, opened 2016-11-23
Hash prefix everywhere

karlbaker02:


src/main/java/uk/gov/register/core/Entry.java
```diff
@@ -17,12 +18,12 @@
...
 public class Entry {
     private final int entryNumber;
-    private final String sha256hex;
+    private final HashValue hashValue;
     private final Instant timestamp;
 
     public Entry(int entryNumber, String sha256hex, Instant timestamp) {
         this.entryNumber = entryNumber;
-        this.sha256hex = sha256hex;
+        this.hashValue = new HashValue(HashingAlgorithm.SHA256, sha256hex);
```
karlbaker02: Perhaps this should be in a factory?

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,21 @@
...
+        this.encodedHash = hashingAlgorithm.toString() + ":" + rawHash;
+    }
+
+    public String encode() {
+        return encodedHash;
+    }
+
+    public String decode() {
+        return rawHash;
+    }
```
michaelabenyohai: The purpose of the decode() method is that it takes an encoded string as a parameter and returns a `HashValue` (as per the example in Trello). i.e. it takes "sha-256:xxxxxx" and returns a HashValue with rawHash of xxxxxx.

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,21 @@
...
+import uk.gov.register.core.HashingAlgorithm;
+
+public class HashValue {
+    private final String rawHash;
+    private final String encodedHash;
+
+    public HashValue(HashingAlgorithm hashingAlgorithm, String rawHash) {
+        this.rawHash = rawHash;
+        this.encodedHash = hashingAlgorithm.toString() + ":" + rawHash;
+    }
```
michaelabenyohai: We don't need to store the encoded hash but compute it on the fly in the encode() method?

karlbaker02: The reason I'm storing the encoded hash is so that it doesn't have to be recomputed on the fly every time it is called.

src/main/java/uk/gov/register/util/HashValue.java
karlbaker02: Say you have a hash `sha-256:d8630e599d42d13a29836f256a75be7be1be60609d34f5040d3095db23153fbb`. Every time you want to decode this hash, if I were to go by the card in Trello, I'd be returning `new HashValue(rawHash)`. To me this seems wasteful in that we're creating a new `HashValue` object when we will already have an exact duplicate in memory (so long as we're decoding a previously encountered hash).

karlbaker02: Additionally, rather than `decode`, the method should be called something like `getRawHash`, because as you mention it isn't decoding anything.

michaelabenyohai:
@karlbaker02 this needs to be rebased

PR #82, opened 2016-11-23
Return Bad Request when a user sends a non-integer value

samcrang:
Previously we would return 500 if the query string contained values that could not be parsed into an Integer; this is not correct as the error lies with the client and should be in the 4xx range. Example: 

```
$ curl -I  "https://country.register.gov.uk/records?page-size=not-a-number"
HTTP/1.1 500 Internal Server Error
...
```

[`IntParam`](http://www.dropwizard.io/0.7.0/dropwizard-jersey/apidocs/io/dropwizard/jersey/params/IntParam.html) is the magic that makes this easy and we should probably use it for `PathParam`s that take integers too.


philandstuff:
:shipit:

PR #83, opened 2016-11-23
Move initial db schema creation into flyway

philandstuff:
We have a home-grown mostly-idempotent script for setting up the initial
state of the database: ensure_schema.sql.  It was managed by a class,
SchemaCreator.

This commit moves that SQL script into Flyway, as a V1 initial schema
script, and deletes the SchemaCreator class.

I tested this locally like this:

  - I ran the app on master, so that I had the schema_version table with
    the V2 script applied in my openregister_java db
  - I then ran the app on this commit, so that I could see what happens
    when we try to run a migration against a database that already has
    V2 applied
  - I then dropped and recreated the database, so that I could see what
    happens when we run a migration against a clean database.

It all works as expected.

EntryMapperTest, which was always a bit of an oddball, had to be changed
because it used SchemaCreator to set up the initial state of the
database.  I fixed this by making it extend FunctionalTestBase and
letting the base class manage the schema creation.  This means that
EntryMapperTest has a full DropwizardAppRule when it doesn't actually
make any requests against the running app, but at least we now have a
consistent way of setting up the database in each test rather than
having one test with its own custom database setup stuff going on.


chr1sdi:
👍

PR #87, opened 2016-11-24
Fix useless header tag + unimplemented exception

chr1sdi:
https://trello.com/c/sUnoWz2L/490-bug-empty-h1-on-every-page

When only `header` was used both TAGs header were pulled on the page regardless of their fragment names, hence the second one page-header was rendered empty where it should not

This also adds a mapper to catch our not implemented/unsupported operation exception and renders dedicated page for it.

### DIFF
--- a/src/main/resources/templates/400.html
+++ b/src/main/resources/templates/400.html
@@ -3,7 +3,7 @@
 <html xmlns:th="http://www.thymeleaf.org" xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
 <head th:include="main.html::head"></head>
 <body>
-<header th:replace="main.html::header"></header>
+<header th:replace="main.html::global-header"></header>
 
 <main id="main" role="main">
     <div th:replace="main.html::phase"></div>
@@ -15,3 +15,4 @@
 <footer th:replace="main.html::footer"></footer>
 </body>
 </html>
+xw
--- a/src/main/resources/templates/400.html
+++ b/src/main/resources/templates/400.html
@@ -15,4 +15,3 @@
 <footer th:replace="main.html::footer"></footer>
 </body>
 </html>
-xw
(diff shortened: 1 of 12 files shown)


src/main/resources/templates/400.html
```diff
@@ -15,3 +15,4 @@
 <footer th:replace="main.html::footer"></footer>
 </body>
 </html>
+xw
```
philandstuff: vim fail

src/main/resources/templates/not-implemented.html
```diff
@@ -3,11 +3,12 @@
...
 <head th:include="main.html::head"></head>
 <body>
-<header th:replace="main.html::header"></header>
+<header th:replace="main.html::global-header"></header>
 
 <main id="main" role="main">
     <div th:replace="main.html::phase"></div>
-    <h1>This page has not been implemented</h1>
+    <a href="/">&lt; back</a>
+    <header th:replace="main.html::page-header(title='This page has not been implemented yet')"></header>
```
philandstuff: this `header` is inside the `main` but on other pages it's outside. is that a problem?

src/main/java/uk/gov/register/providers/ThrowableExceptionMapper.java
```diff
@@ -35,7 +35,15 @@ public Response toResponse(Throwable exception) {
             return Response.status(Response.Status.NOT_FOUND)
                     .header(HttpHeaders.CONTENT_TYPE, ExtraMediaType.TEXT_HTML)
                     .entity(viewFactory.thymeleafView("404.html"))
-                    .build();        }
+                    .build();
+        }
+
+        if (exception instanceof UnsupportedOperationException) {
```
philandstuff: don't like this 👎  there is library code that can throw `UnsupportedOperationException`

PR #88, opened 2016-11-24
Hash prefix everywhere #2

karlbaker02:
Rebased with latest master.


src/main/java/uk/gov/register/core/Entry.java
```diff
@@ -53,7 +53,7 @@ public Integer getEntryNumber() {
 
     @JsonProperty("item-hash")
     public String getItemHash() {
-        return "sha-256:" + sha256hex;
+        return hashValue.toString();
```
michaelabenyohai: We can now remove this method entirely and put `@JsonProperty("item-hash")` and `@JsonSerialize(using = ToStringSerializer.class)` on the `getSha256hex()` method. Jackson should then call `toString()` behind the scenes.

src/main/java/uk/gov/register/core/Entry.java
```diff
@@ -80,13 +80,14 @@ public boolean equals(Object o) {
...
 
         if (entryNumber != entry.entryNumber) return false;
-        return sha256hex == null ? entry.sha256hex == null : sha256hex.equals(entry.sha256hex);
+        return hashValue == null ? entry.hashValue == null : hashValue.getValue().equals(entry.hashValue.getValue());
     }
 
     @Override
     public int hashCode() {
-        int result = sha256hex != null ? sha256hex.hashCode() : 0;
+        String hash = hashValue.getValue();
```
michaelabenyohai: Similarly, you should not use `hashValue.getValue()` since you have implemented `HashValue.hashCode()`.

src/main/java/uk/gov/register/core/EntryLog.java
```diff
@@ -73,15 +73,21 @@ public RegisterProof getRegisterProof(int entryNo) {
     public EntryProof getEntryProof(int entryNumber, int totalEntries) {
         List<String> auditProof = backingStoreDriver.withVerifiableLog(verifiableLog ->
                 verifiableLog.auditProof(entryNumber, totalEntries)
-                        .stream().map(this::bytesToString).collect(Collectors.toList()));
+                        .stream()
+                        .map(this::bytesToString)
+                        .map(p -> HashingAlgorithm.SHA256.toString() + ":" + p)
```
michaelabenyohai: You should not need to do this here if you change the `EntryProof.auditPath` property `List<HashValue>`. One purpose of creating the `HashValue` class is that we do not concatenate the "sha-256:" prefix in multiple places in code.

src/main/java/uk/gov/register/core/EntryLog.java
```diff
@@ -73,15 +73,21 @@ public RegisterProof getRegisterProof(int entryNo) {
...
         return new EntryProof(Integer.toString(entryNumber), auditProof);
     }
 
     public ConsistencyProof getConsistencyProof(int totalEntries1, int totalEntries2) {
         List<String> consistencyProof = backingStoreDriver.withVerifiableLog(verifiableLog ->
                 verifiableLog.consistencyProof(totalEntries1, totalEntries2))
-                .stream().map(this::bytesToString).collect(Collectors.toList());
+                .stream()
+                .map(this::bytesToString)
+                .map(p -> HashingAlgorithm.SHA256.toString() + ":" + p)
```
michaelabenyohai: Same as above. `ConsistencyProof.consistencyNodes` should be changed to be of type `List<HashValue>` and that will mean you will not need to repeat string concatenation here.

src/main/java/uk/gov/register/core/Item.java
```diff
@@ -73,14 +72,15 @@ public boolean equals(Object o) {
...
-        if (sha256hex != null ? !sha256hex.equals(item.sha256hex) : item.sha256hex != null) return false;
+        if (hashValue != null ? !hashValue.equals(item.hashValue) : item.hashValue != null) return false;
         return content != null ? content.equals(item.content) : item.content == null;
 
     }
 
     @Override
     public int hashCode() {
-        int result = sha256hex != null ? sha256hex.hashCode() : 0;
+        String hash = hashValue.getValue();
```
michaelabenyohai: You should not use `hashValue.getValue().getHashCode()` since you have already implemented `hashValue.getHashCode()`. You should just need to use `hashValue.getHashCode()`.

src/main/java/uk/gov/register/resources/ItemResource.java
```diff
@@ -25,14 +27,14 @@ public ItemResource(RegisterReadOnly register, ViewFactory viewFactory) {
     @Path("/{item-hash}")
     @Produces({ExtraMediaType.TEXT_HTML, MediaType.APPLICATION_JSON, ExtraMediaType.TEXT_YAML, ExtraMediaType.TEXT_CSV, ExtraMediaType.TEXT_TSV, ExtraMediaType.TEXT_TTL})
     public AttributionView getItemByHex(@PathParam("item-hash") String itemHash) {
-        String sha256Regex = "(sha-256:)(.*)";
+        String sha256Regex = "(" + HashingAlgorithm.SHA256.toString() + ":)(.*)";
```
michaelabenyohai: This is an example of a place where you should be using `HashValue.decode(encodedHash)` as the `itemHash` value is an example of an encoded hash that needs to be decoded. The regex logic could even be used in the `HashValue.decode(encodeHash)` method.

src/main/java/uk/gov/register/resources/ItemResource.java
```diff
@@ -25,14 +27,14 @@ public ItemResource(RegisterReadOnly register, ViewFactory viewFactory) {
...
         if (itemHash.matches(sha256Regex)) {
             return getItemBySHA256(itemHash.replaceAll(sha256Regex, "$2"));
         }
         throw new NotFoundException("No item found with item hex: " + sha256Regex);
     }
 
     private ItemView getItemBySHA256(String sha256Hash) {
-        return register.getItemBySha256(sha256Hash).map(viewFactory::getItemView).orElseThrow(NotFoundException::new);
+        return register.getItemBySha256(new HashValue(HashingAlgorithm.SHA256, sha256Hash)).map(viewFactory::getItemView).orElseThrow(NotFoundException::new);
```
michaelabenyohai: you shouldn't need to `new` a `HashValue` if you use `decode()`

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -75,17 +80,18 @@ public void addCommand(String s) {
                 break;
             case "append-entry":
                 if (parts.length == 4) {
-                    Entry entry = new Entry(0, stripPrefix(parts[2]), Instant.parse(parts[1]), parts[3]);
+                    Entry entry = new Entry(0, new HashValue(HashingAlgorithm.SHA256, stripPrefix(parts[2])), Instant.parse(parts[1]), parts[3]);
```
michaelabenyohai: You should be able to get rid of `stripPrefix` and instead use `decode()` - the benefit of the new `HashValue` object is that all this prefix removal happens in one place (i.e. `decode()`).

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -75,17 +80,18 @@ public void addCommand(String s) {
...
                     entries.put(position++, entry);
-                    updateItemHashCount(entry.getSha256hex());
+                    updateItemHashCount(entry.getSha256hex().getValue());
```
michaelabenyohai: You should not need to use `.getvalue()` here. Can we change `itemHashToEntryCount` to be a `HashMap<HashValue, Integer>`. If you have implemented `HashValue.hashCode()` correctly this should work the same.

src/main/java/uk/gov/register/serialization/CommandParser.java
```diff
@@ -75,17 +80,18 @@ public void addCommand(String s) {
...
                 } else {
                     LOG.error("append entry line must have 4 elements, was : " + s);
                     throw new SerializedRegisterParseException("append entry line must have 4 elements, was : " + s);
                 }
                 break;
             case "assert-root-hash":
                 if (parts.length == 2) {
-                    RegisterProof registerProof = new RegisterProof(parts[1]);
+                    HashValue hash = HashValue.decode(HashingAlgorithm.SHA256.toString(), parts[1]);
+                    RegisterProof registerProof = new RegisterProof(hash.getValue());
```
michaelabenyohai: `RegisterProof` should take a `HashValue` in its constructor rather than a `String` and `RegisterProof.rootHash` property should be of type `HashValue`.

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new RuntimeException("Hash cannot be decoded");
+        }
+
+        String[] parts = encodedHash.split(hashingAlgorithm + ":");
+
+        if (parts.length != 2) {
+            throw new RuntimeException(String.format("Cannot create HashValue from encoded hash %s", encodedHash));
+        }
+
+        return new HashValue(HashingAlgorithm.SHA256, parts[1]);
```
michaelabenyohai: Why is HashingAlgorithm hardcoded here but nowhere else?

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,55 @@
...
+        if (!encodedHash.startsWith(hashingAlgorithm)) {
+            throw new RuntimeException("Hash cannot be decoded");
+        }
+
+        String[] parts = encodedHash.split(hashingAlgorithm + ":");
+
+        if (parts.length != 2) {
+            throw new RuntimeException(String.format("Cannot create HashValue from encoded hash %s", encodedHash));
+        }
+
```
michaelabenyohai: You do not check for the situation where you get `sha-256abcde:fghijk` - in this case parts[1] would be `fghijk` when this should not have parsed. I think you should check that the `encodeHash` starts with `sha-256:` rather than checking that it starts with `sha-256`.

Also can you throw a more specific exception rather than `RuntimeException`? I think it should also be a checked exception (rather than unchecked) so that we know in any code that calls this that we need to deal with the situation where the encodedHash cannot be decoded.

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || o.getClass() != this.getClass()) return false;
+
+        HashValue that = (HashValue) o;
+
+        return this.getValue().equals(that.getValue());
+    }
```
michaelabenyohai: Your equals method should take into account the `hashingAlgorithm` property and check that its equal between the two objects, since it is configurable. Either that or hard code `hashingAlgorithm` to always be sha-256 (sice it is always that anyway).

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public int hashCode() {
+        return 31 * getValue().hashCode();
+    }
```
michaelabenyohai: Same as above, `hashCode()` should check for differences in the `hashingAlgorithm` property.

src/test/java/uk/gov/register/util/HashValueTest.java
```diff
@@ -0,0 +1,52 @@
+package uk.gov.register.util;
+
+import org.junit.Test;
+import uk.gov.register.core.HashingAlgorithm;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+public class HashValueTest {
```
michaelabenyohai: Can we add tests for equals and hashcode for when the `hashingAlgorithm` is different?

Also can we add tests for cases where decode will fail and the exceptions we expect?

src/main/java/uk/gov/register/util/HashValue.java
karlbaker02: `sha-256abcde:fghijk` would be picked up by the fact that `encodedHash.split(hashingAlgorithm + ":")` would not be able to split the string.

Point taken on checked exceptions.

src/main/java/uk/gov/register/core/Entry.java
karlbaker02: That's fine when we're referencing `item-hash`, but if we ever call `getSha256hex()` from code we're returning a `HashValue` object, as opposed to a `String`, so we'll need to explicitly call `toString()` if we want to compare this `HashValue` to a `String`.

src/main/java/uk/gov/register/views/ConsistencyProof.java
```diff
@@ -23,7 +26,9 @@ public String getProofIdentifier() {
 
     @SuppressWarnings("unused, used as jsonproperty")
     @JsonProperty("merkle-consistency-nodes")
-    public List<String> getConsistencyNodes() {
+    public List<HashValue> getConsistencyNodes() {
+        List<String> nodes = consistencyNodes.stream().map(p -> p.toString()).collect(Collectors.toList());
```
michaelabenyohai: The `nodes` variable isn't being used and isn't required.

Also this results in `"merkle-consistency-nodes":[{"value":"bbb6942aa14ed87aeb180e9f9bfdeab62bc537222dae0f8ee91433c92c7de48d"}` being returned incorrectly by the endpoint.

If you annotate this method with `@JsonSerialize(using = ToStringSerializer.class)` then you will get the correct result.

src/main/java/uk/gov/register/views/EntryProof.java
```diff
@@ -31,7 +31,7 @@ public String getEntryNumber() {
 
     @SuppressWarnings("unused, used as jsonproperty")
     @JsonProperty("merkle-audit-path")
-    public List<String> getAuditPath() {
+    public List<HashValue> getAuditPath() {
```
michaelabenyohai: This method needs to be annotated with `@JsonSerialize(using = ToStringSerializer.class)`. This will correct the problem where the endpoint is currently returning `"merkle-audit-path":[{"value":"cc3c7f8f149fb415152a46d0c35098bba6644271402e8c68138e8b0328848ac0"}]`

michaelabenyohai: Please also update either `VerifiableLogResourceTest` or `VerifiableLogResourceFunctionalTest` to add a test to make sure the correct prefix is added to the hashes in the `merkle-audit-path`.

src/main/java/uk/gov/register/views/ConsistencyProof.java
michaelabenyohai: Please also update either `VerifiableLogResourceTest` or `VerifiableLogResourceFunctionalTest` to add a test to make sure the correct prefix is added to the hashes in the `merkle-consistency-nodes`.

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,65 @@
...
+        this.hashingAlgorithm = hashingAlgorithm.toString();
+        this.value = value;
+    }
+
+    public HashValue(String hashingAlgorithm, String value) {
+        this.hashingAlgorithm = hashingAlgorithm;
+        this.value = value;
+    }
+
+    @JsonValue
```
michaelabenyohai: `@JsonValue` should not be required here. You have used `@JsonSerialize(using = ToStringSerializer.class)` in the relevant places so it is using the `toString()` method anyway (which points to `encode()`).

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,65 @@
...
+    public String encode() {
+        return hashingAlgorithm + ":" + value;
+    }
+
+    @JsonIgnore
```
michaelabenyohai: For the same reason you should not need `@JsonIgnore` here.

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || o.getClass() != this.getClass()) return false;
+
+        HashValue that = (HashValue) o;
+
+        return this.encode().equals(that.encode());
```
michaelabenyohai: `equals` should not be based on a method as the method could be changed by someone who doesn't notice that `equals` is using it. It needs to be based on the `value` and `hashingAlgorithm` properties themselves. See other examples in `Item.equals` and `Entry.equals`

src/main/java/uk/gov/register/util/HashValue.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * encode().hashCode();
```
michaelabenyohai: Same as for `equals`, `hashCode` should not be based on a method but directly on the `value` and `hashingAlgorithm` properties.

PR #89, opened 2016-11-25
reinstate RepresentationsFunctionalTest

philandstuff:
This reinstates the RepresentationsFunctionalTest, which was ignored in
20e1597a7a7e9d97b.

I achieved this by removing the `FunctionalTestBase` superclass and
using a separate DropwizardAppRule which sets the `register` value to
`register`.

Furthermore, now that we have the `/load-rsf` endpoint, we can load data
in as RSF rather than using the grubby DBSupport class to write directly
into the database tables.


chr1sdi:
👍  merge! :rage4:

PR #90, opened 2016-11-25
Restart Docker app containers that haven't explicitly been stopped

samcrang:
If the container dies then we can have the Docker service automatically restart the container. We should see less downtime when apps die in environments where there is little or no redundancy.


chr1sdi:
👍

samcrang:
cheers m8

PR #95, opened 2016-11-28
migrate database outside of the app itself

philandstuff:
This changes the app so that it doesn't migrate its own database on app
startup.  This means that whenever the app runs, something else needs to
ensure the database has been migrated appropriately.  There are four
situations:

 - for actual deployment, the start-service.sh script runs the
   migration (using a command I manually tested on an existing EC2 box)
 - for ./run-application, we have an extra gradle task migrateMainDb
 - for functional tests, RegisterRule migrates the database before
   running the tests
 - for conformance tests, there was already the appropriate usage of
   {create,drop}_schema.sh, but the waitForApplicationToStart()
   operation broke because the app didn't return a 200 response on the
   application port when the database hadn't been migrated already.  I
   replaced this with a call to /healthcheck on the admin port instead.

The longer-term goal is to entirely separate application deployment from
database deployment, so that we can have automatic rollbacks for apps
and have a separate deployment pipeline for database migrations.

### DIFF
--- a/src/test/java/uk/gov/register/functional/app/RegisterRule.java
+++ b/src/test/java/uk/gov/register/functional/app/RegisterRule.java
@@ -70,7 +70,6 @@ public class RegisterRule implements TestRule {
     private void migrateDb() {
         RegisterConfiguration configuration = appRule.getConfiguration();
         Flyway flyway = configuration.getFlywayFactory().build(configuration.getDatabase().build(appRule.getEnvironment().metrics(), "flyway_db"));
-        flyway.setBaselineVersionAsString("0");
         flyway.migrate();
     }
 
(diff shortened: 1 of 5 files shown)


michaelabenyohai:
@philandstuff please can you rebase?

philandstuff:
Rebased, but hold off merging for the moment please

philandstuff:
Ready for review + merge now!

src/test/java/uk/gov/register/functional/app/RegisterRule.java
```diff
@@ -52,13 +48,30 @@ public Statement apply(Statement base, Description description) {
...
+        authenticatedTarget.register(HttpAuthenticationFeature.basicBuilder().credentials("foo", "bar").build());
+
+        migrateDb();
+        wipe();
+    }
+
+    private void migrateDb() {
+        RegisterConfiguration configuration = appRule.getConfiguration();
+        Flyway flyway = configuration.getFlywayFactory().build(configuration.getDatabase().build(appRule.getEnvironment().metrics(), "flyway_db"));
+        flyway.setBaselineVersionAsString("0");
```
michaelabenyohai: Can we remove this? It doesn't appear to be set anywhere outside of the app e.g. from run-application.sh.

philandstuff:
@chr1sdi can you review 4e859ca in particular?

michaelabenyohai:
I'm happy. @chr1sdi please can you take a look at whether the last commit is ok?

chr1sdi:
👍

PR #97, opened 2016-11-29
Enhance all the things - make sure that GA events are fired 

chr1sdi:
when navigating to external links or downloading the data on all browsers

The problem was that firefox was redirecting before the ga event was sent. The solution is to cancel browser default action, send GA event and then continue (with timeout on GA event). Google analytics: 
https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits#knowing_when_the_hit_has_been_sent

https://trello.com/c/8jMQNDqA/488-bug-google-analytics-events-for-download-page-aren-t-tracked-in-firefox-exception-handling

### DIFF
--- a/src/main/resources/assets/js/analytics.js
+++ b/src/main/resources/assets/js/analytics.js
@@ -1,32 +1,74 @@
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-})(window,document,'script','//www.google-analytics.com/analytics.js','ga');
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/js/analytics.js
```diff
@@ -1,32 +1,74 @@
...
+
+  sendEvent = function(category, action, targetLabel, fnHitCallback){
+    checkTrackerAndExecute(function(){
+      callbackExecuted = false;
+      var executeHitCallback = function () {
+        if (!callbackExecuted) {
+          callbackExecuted = true;
+          fnHitCallback();
+        }
+      }
```
karlbaker02: Missing ;

karlbaker02:
👍

philandstuff:
i suspect that `callbackExecuted = false;` could be `var callbackExecuted = false;` because the executeHitCallback function closes over it.

chr1sdi:
@philandstuff solved in https://github.com/openregister/openregister-java/pull/102/files#diff-241e835ebefcc5f1895770827405c019R22

PR #101, opened 2016-11-30
Papercuts

symroe:
Couple of easy fixes. Not sure about the vars in 423bab0 – see the comment on that commit for more. Happy to fixup the commit with the correct variables and words if need be.

### DIFF
--- a/src/main/resources/assets/sass/main.scss
+++ b/src/main/resources/assets/sass/main.scss
@@ -139,7 +139,7 @@ main {
 .field-list li {
   font-family: monospace;
   font-size: 16px;
-  display: inline-block;
+  display: inline;
 }
 
 .field-list li:after {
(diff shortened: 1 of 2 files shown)


philandstuff:
I spoke to @symroe IRL about this. The `table-cell` trick doesn't seem to quite fix the comma whitespace issue, and the "view this register" link needs to, at least, be configurable so it doesn't appear on registers other than the register register.

We can talk further about next steps.

michaelabenyohai:
https://github.com/openregister/openregister-java/pull/108 addresses the "view this register" link and configures it only for the register register.

michaelabenyohai:
Moved table-cell trick to its won branch - turns out chrome was caching CSS and it does work locally for me now. New PR at https://github.com/openregister/openregister-java/pull/111.

philandstuff:
Going to close as I think we've dealt with the issues in this PR. @symroe please reopen if there are still things we haven't addressed.

PR #107, opened 2016-12-01
Delete data from register

chr1sdi:
https://trello.com/c/chlAXfBX/484-create-endpoint-to-delete-data-from-a-register
to test this add `enableRegisterDataDelete: true` to `config.yaml`

### DIFF
--- a/src/main/java/uk/gov/register/RegisterApplication.java
+++ b/src/main/java/uk/gov/register/RegisterApplication.java
@@ -16,6 +16,7 @@ import io.dropwizard.jersey.setup.JerseyEnvironment;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import io.dropwizard.views.ViewBundle;
+import org.flywaydb.core.Flyway;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 import org.skife.jdbi.v2.DBI;
 import uk.gov.organisation.client.GovukOrganisationClient;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/register/RegisterApplication.java
```diff
@@ -115,6 +115,7 @@ public void run(RegisterConfiguration configuration, Environment environment) th
         jersey.register(new AbstractBinder() {
             @Override
             protected void configure() {
+                bind(flyway).to(Flyway.class);
```
philandstuff: this will conflict with #95, which removes the `flyway` variable

src/test/java/uk/gov/register/functional/app/RegisterRule.java
```diff
@@ -81,6 +81,10 @@ public Response getRequest(String path) {
         return target().path(path).request().get();
     }
 
+    public Response deleteRequest(String path) {
```
philandstuff: does it make sense to make this specific to the only DELETE endpoint that we have?

src/main/java/uk/gov/register/RegisterApplication.java
```diff
@@ -99,7 +99,9 @@ public void run(RegisterConfiguration configuration, Environment environment) th
         DBI jdbi = dbiFactory.build(environment, configuration.getDatabase(), "postgres");
 
         Flyway flyway = configuration.getFlywayFactory().build(configuration.getDatabase().build(environment.metrics(), "flyway_db"));
-
+        flyway.setBaselineVersionAsString("0");
+        flyway.migrate();
```
philandstuff: Can we not do this please? The whole point of #95 is to remove this call to `flyway.migrate()`

src/main/java/uk/gov/register/RegisterApplication.java
chr1sdi: no worries, will rebase

src/test/java/uk/gov/register/functional/app/RegisterRule.java
chr1sdi: no worries, will look at this

src/main/java/uk/gov/register/RegisterApplication.java
chr1sdi: no worries, this had to be brought back because #95 wasn't merged at that time, now it will disappear :)

src/main/java/uk/gov/register/resources/HomePageResource.java
```diff
@@ -51,4 +58,15 @@ public String analyticsTrackingId() {
                 trackingId -> "var gaTrackingId = \"" + trackingId + "\";\n"
         ).orElse("");
     }
+
+    @DELETE
+    @PermitAll
+    @Path("/delete-register-data")
+    @DataDeleteNotAllowed
+    public Response deleteRegisterData() {
```
philandstuff: I wonder if this makes sense to be a separate class, rather than shoving it into the HomePageResource?

PR #113, opened 2016-12-01
Introduce resolution abstractions

philandstuff:
This introduces two new abstractions:

 - LinkResolver, which resolves a LinkValue into a URI
 - RegisterResolver, which gives the base URI of a register

At the moment, the only method we have for doing this is URI templating.  However, we should collect all instances of URL templating so that, should we change how we discover other registers, we can change it in one place only.

See individual commit messages for more details.


karlbaker02:
Looks good.

PR #115, opened 2016-12-02
support multiple registers

philandstuff:
This introduces a new configuration item `otherRegisters`, which
allows you to specify a map from register names to database
configurations for other registers that this app can serve.

The app decides which register to use on a per-request basis, by
inspecting the X-Forwarded-Host or Host headers, and pulling out the
bit before the first dot.  If it matches a named register in
`otherRegisters`, it serves from that register; otherwise, it serves
from itself.

To do this, there is a new class AllTheRegisters, which allows you to
pick out a register using its getRegisterByName() method.  This
returns an EverythingAboutARegister instance.
EverythingAboutARegister has all the things you need for an individual
register instance - MemoizationStore, DBI, RegisterMetadata, etc.

I also removed RegisterNameConfiguration, as that is no longer a
globally unique thing.  You now need to inject
EverythingAboutARegister or RegisterData and get the name from that.

All this code was just a first attempt and a proof-of-concept.  Here
are some outstanding questions:

 - what's our strategy to move forward with this?  do we want to
   deploy this to multiple environments?
 - do we want to stick with the "default register" model? I did this
   so that we could make a forwards-compatible change to the
   configuration file format, but should we ensure that each register
   is explicitly called?
 - have we got the right abstractions here?  Is
   EverythingAboutARegister really duplicating the responsibility that
   PostgresRegister already has?  Should those two classes be merged?
 - what's going on with all this dependency injection stuff? Can we
   simplify it, by getting rid of some of the `Factory<>`s and
   `Provider<>`s?

I've tested this locally, manually, but there are some bits I haven't
tested:

 - I haven't tested proofs, to ensure that the MemoizationStores are
   truly isolated from each other

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -204,13 +204,7 @@ task createDeployableBundle(type: Zip) {
     destinationDir file('deployable_bundle') // directory that you want your archive to be placed in
 }
 
-task migrateMainDb(type: JavaExec, dependsOn: ['classes']) {
-    main = 'uk.gov.register.RegisterApplication'
-    classpath = sourceSets.main.runtimeClasspath
-    args = ["db", "migrate", "config.yaml"]
-}
-
-task(run, type: JavaExec, dependsOn: ['classes', 'migrateMainDb']) {
+task(run, type: JavaExec, dependsOn: ['classes']) {
     main = 'uk.gov.register.RegisterApplication'
     classpath = sourceSets.main.runtimeClasspath
     args = ["server", "config.yaml"]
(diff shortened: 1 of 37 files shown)


andyjpb:
Can we match the whole Host / X-Forwarded-Host header so that we can support multiple discovery environments or multiple environments from a single system?

philandstuff:
@andyjpb can you give an example of what you mean? I'm not sure I understand.

philandstuff:
@michaelabenyohai mentioned that we need to ensure the flyway migration script still works, which currently I think it doesn't.

philandstuff:
not ready for merging. closing for the moment

PR #116, opened 2016-12-05
Use batching for inserts. Keep limited debug statements. Do not write…

johnollier:
… staged changes before reading entries count.

The new register transaction code means that after parsing the RSF into commands all item and entry objects are held in memory and after validation they are written to the database. We run out of memory if the collections of items/entries are inserted as a single batch. Adding the BatchChunkSize annotation allows the insertion to be done in batches within a single transaction.

When Entries are created we need to calculate the entry number which is the maximum entry number of existing entries either in the database or in memory. Previously we wrote all staged items/entries to the database before reading the maximum entry number. This resulted in an interaction with the DB every time. I've added a method to read from the db without writing staged item/entries first and used that to get the total number of entries.

### DIFF
--- a/src/main/java/uk/gov/register/store/postgres/PostgresDriverNonTransactional.java
+++ b/src/main/java/uk/gov/register/store/postgres/PostgresDriverNonTransactional.java
@@ -57,11 +57,6 @@ public class PostgresDriverNonTransactional extends PostgresDriver {
     }
 
     @Override
-    protected <ReturnType> ReturnType withHandleRead(HandleCallback<ReturnType> callback) {
-        return dbi.withHandle(callback);
-    }
-
-    @Override
     protected <ReturnType> ReturnType inTransaction(HandleCallback<ReturnType> callback) {
         return dbi.inTransaction((handle, status) -> callback.withHandle(handle));
     }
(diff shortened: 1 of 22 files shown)


src/main/java/uk/gov/register/store/postgres/PostgresDriverNonTransactional.java
```diff
@@ -57,6 +57,11 @@ protected void useHandle(HandleConsumer callback) {
     }
 
     @Override
+    protected <ReturnType> ReturnType withHandleRead(HandleCallback<ReturnType> callback) {
+        return dbi.withHandle(callback);
+    }
```
michaelabenyohai: Good spot on this bug! It certainly wasn't meant to be going to the database for every call to `getItemBySha256()`.

I'm not sure about adding a `withHandleRead()` though, since it doesn't make much sense for `PostgresDriverNonTransactional`.

Could we instead add something like a `committedEntryCount` property to the `PostgresDriverTransactional` which we can initialise to `super.getTotalEntries()` in the constructor. Then for `PostgresDriverTransactional.getTotalEntries()` we could return `committedEntryCount + stagedItems.size()`. We'd also have to reset `committedEntryCount` on the `writeEntries()` method too. What do you think?

PR #118, opened 2016-12-05
Remove another instance of flyway baseline

philandstuff:
For consistency with 681d234e1b7bddd03f356ddab80b367de3626702 - we
should configure flyway the same way everywhere.


chr1sdi:
👍

chr1sdi:
very nice

PR #119, opened 2016-12-06
Change history page to point at shool history page and not country

chr1sdi:
Since we have it may as well update history page for our dev config - usually everything is for schools.


philandstuff:
:yolo:

PR #120, opened 2016-12-06
Use RegisterReadOnly interface, not RegisterData, in views

philandstuff:
This migrates the views to use the RegisterReadOnly interface instead
of RegisterData.  It also adds getRegisterName() and
getRegisterMetadata() to that interface.

My idea here is that we can start to migrate other things to the
RegisterReadOnly interface - like getting the custodian and the
branding.

This also will help with the work I'm doing towards multiple registers
in one app server.

### DIFF
--- a/src/main/java/uk/gov/register/RegisterApplication.java
+++ b/src/main/java/uk/gov/register/RegisterApplication.java
@@ -24,6 +24,7 @@ import uk.gov.register.auth.AuthBundle;
 import uk.gov.register.configuration.FieldsConfiguration;
 import uk.gov.register.configuration.PublicBodiesConfiguration;
 import uk.gov.register.configuration.RegisterFieldsConfiguration;
+import uk.gov.register.configuration.RegisterNameConfiguration;
 import uk.gov.register.configuration.RegistersConfiguration;
 import uk.gov.register.core.PostgresRegister;
 import uk.gov.register.core.Register;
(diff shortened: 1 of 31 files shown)


johnollier:
Just a very general point - when using MVC frameworks, I've always assumed that the objects to be passed into a view should be 'beans' as opposed to 'services'. I had a quick look at the DropWizard docs and in their simple example a View class just wraps a Domain class. 

The implementation of RegisterReadOnly which gets passed to the views is PostgresRegister which is a mixture of both. The 'service' methods are not being called in the view but it would be possible to do so. Is that a good thing. When we go to multi-tenancy will the Register Metadata for the request get looked up in the view?

philandstuff:
Yeah, I'm not sure either.  I think we're in a funny situation because we have a single View that gets rendered to different serialisations, which have different requirements.

In particular, right now you have to provide a Registry and a Branding to each subclass of AttributionView when it's created, and you get these by calling the GOV.UK organisations API.  But they only get used by the HTML representation - not by any of the others - so we're paying the price of accessing the API even when we don't need to.

We could save this cost by making the View responsible for getting the data, on-demand, from the API, so that the API call is only made when the Registry/Branding is needed.

So.. while in principle I agree that views should just be bags of data; I also disagree.  And I think some of these problems would go away if we moved away from using Dropwizard for HTML/CSS/JS stuff.

philandstuff:
going to go away and think about what I've done

PR #123, opened 2016-12-06
Rsf performance

johnollier:
Rewrite to read maximum entry number at start of transaction and use for entry creation if there are no staged entries.

### DIFF
--- a/src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
+++ b/src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
@@ -28,7 +28,7 @@ public class PostgresDriverTransactional extends PostgresDriver {
     private final Handle handle;
 
     private final List<Entry> stagedEntries;
-    private final HashMap<HashValue,Item> stagedItems;
+    private final HashMap<HashValue, Item> stagedItems;
     private final HashMap<String, Integer> stagedCurrentKeys;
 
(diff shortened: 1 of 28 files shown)


src/main/java/uk/gov/register/store/postgres/PostgresDriverTransactional.java
```diff
@@ -38,18 +39,21 @@ private PostgresDriverTransactional(Handle handle, TransactionalMemoizationStore
         this.stagedEntries = new ArrayList<>();
         this.stagedItems = new HashMap<>();
         this.stagedCurrentKeys = new HashMap<>();
+        this.initialEntryCount = super.getTotalEntries();
```
michaelabenyohai: `initialEntryCount` also needs to be reset whenever we do `writeEntries()` (after the `isEmpty()` check). Currently, if we flush the staged entries half way through a transaction (e.g. if there is an assert-root-hash half way through the file) then we will return 0 from `getTotalEntries()` even though there are entries in the database.

Maybe we should call it something like `committedEntryCount` since it can change after being initialised?

michaelabenyohai:
Can we also create some tests to detect the original bug - e.g. test how many times it makes a call to the database.

johnollier:
Maximum committed entry number is now reset on write to db. Added a test to verify the number of db interactions when adding entries.

PR #124, opened 2016-12-06
Remove key from all data download formats for Record

karlbaker02:


src/main/java/uk/gov/register/views/representations/CsvWriter.java
```diff
@@ -27,6 +28,7 @@ public void writeTo(CsvRepresentationView view, Class<?> type, Type genericType,
         CsvRepresentation csvRepresentation = view.csvRepresentation();
 
         objectMapper.writerFor(csvRepresentation.contentType)
+                .with(JsonGenerator.Feature.IGNORE_UNKNOWN)
```
karlbaker02: It appears that the problem this line used to address (mapping an unknown field of `key`) no longer exists, so I'll undo this change.

chr1sdi:
👍  and what @michaelabenyohai said :)

PR #126, opened 2016-12-07
Change the way we manage Item views

philandstuff:
This is a proof-of-concept for how we might rationalise the way we
handle views in our resources.

There are some problems with our views:

 - there's a lot of inheritance, and a lot stuff is passed through
   constructors in order to be added to the superclasses
   AttributionView and ThymeleafView
 - there's quite a bit of behaviour in views -- such as ItemView
   having an ItemConvertor to translate an Item into a data structure
   suitable for rendering
 - we have to do the maximum work just in case our view will be
   rendered as any format.  In particular, we always fetch the GOV.UK
   organisation API data, even though that's only consumed by the HTML
   views

This commit does a few specific things to address these problems in
the specific case of ItemResource.  If we like it we can extend this
to the other resources.

 - reducing inheritance:
   - CsvRepresentationView is converted from an abstract class to an
     interface, which reduces the depth of inheritance
   - AttributionView is changed to a generic class, so that you add to
     it by setting the generic parameter rather than extending
     it.  (The old behaviour is preserved in OldAttributionView for
     the other views outside of ItemResource)
 - reducing behaviour:
   - I've split out separate methods for @Produces(TEXT_HTML) and
     @Produces(everything else).  Jersey understands this and does the
     right thing.  This means we only need to fetch GOV.UK
     organisation API data in the TEXT_HTML method.

The ItemResource has got a little more complex, but ItemView is much
simpler, and the tests have become simpler too, so this feels like a
good change.


philandstuff:
I've discovered a slight wrinkle with splitting up the resource methods between different media types like this: whereas previously, Jersey would pick the first mediatype in the `@Produces` annotation as the default, now it will pick the first mediatype in a randomly-selected method.  This means that if you request `/item/sha-256:fa1afe1...abcde` with the header `Accept: */*`, you might end up with JSON instead of HTML. 

This affects cURL, and some of our tests, but doesn't affect browsers because they specify `Accept: text/html` (well, something a bit more complicated than this but near enough the same).  I suspect this is good enough?

philandstuff:
I've pushed commits refactoring the rest of the views, not just Item.

michaelabenyohai:
@philandstuff please can you rebase

philandstuff:
@michaelabenyohai done

PR #127, opened 2016-12-08
Split EntryLog into separate implementations

philandstuff:
This turns the EntryLog class into an interface, with three separate
implementations:

 - OnDemandEntryLog, for doing one-by-one operations (read-only)
 - TransactionalEntryLog, for doing transactional operations,
   including appends
 - InMemoryEntryLog, a stub implementation for tests

TransactionalEntryLog has swallowed a lot of the functionality from
PostgresDriverTransactional.

OnDemandEntryLog has swallowed a lot of the functionality from
PostgresDriverNonTransactional.


karlbaker02:
Walked through with Phil, all looks good.

PR #128, opened 2016-12-09
Split item store functionality out from PostgresDriver

philandstuff:
Similar to #127 (merged in d6507f0)

This extracts the ItemStore functionality out from BackingStoreDriver
and its subclasses, in a similar way to that which was already done
for EntryLog.  That is, ItemStore has been converted to an interface,
with separate implementations for on-demand and transactional work.

Extra things done:

 - the InMemory*DAOs have been simplified - it makes sense to have one
   class that implements both the query and command interfaces.
 - the InMemoryEntryLog was so similar to the OnDemandEntryLog that
   they could actually share most of their code.
 - there's no need for an InMemoryItemStore -- it's just an
   OnDemandItemStore with InMemoryItemDAO backing it
 - there's some gnarly bits in ItemStore where the getIterator()
   methods need to know those items which are referenced by particular
   entries; so the ItemStore needs access to the entry storage as well
   as the item storage.  This is reflected in how we end up
   constructing the various ItemStore implementations
 - the previously @Ignored test in PostgresDriverTransactionalTest has
   been reinstated in PostgresDriverTransactionalFunctionalTest, in
   terms of a transaction on a PostgresRegister.


philandstuff:
@karlbaker02 I pushed fixes for the things you mentioned today

karlbaker02:
👍

PR #129, opened 2016-12-12
Add handler and url connection for s3 urls. Use urls for reading fiel…

johnollier:
Just for discussion rather than merging at present. But I was interested to see what was required in order to use a 's3://...' URL to point to the config. It involves setting the 'java.protocol.handler.pkgs' property.


philandstuff:
this needs a rebase now FYI

philandstuff:
this is neat! means we can naturally use `s3://` URLs or plain old `file://` or `https://` ones.

PR #133, opened 2016-12-13
Check whether iterator is already closed

michaelabenyohai:
Sometimes the iterator may not be null but may be already closed. This
occurs because we have used the EntryIteratorDAO previously to read up to the end
of the iterator (which will close it). This now checks for the closed iterator and
recreates it if necessary.


philandstuff:
Nice find!

Is it possible to write a test to capture this failure mode?

michaelabenyohai:
Yes I realise there are no tests for this - I will write some!

PR #134, opened 2016-12-14
Fix iterator closed bug

michaelabenyohai:
Sometimes the iterator may not be null but may be already closed. This occurs because we have used the EntryIteratorDAO previously to read up to the end of the iterator (which will close it). This now checks for the closed iterator and recreates it if necessary.


johnollier:
Looks good. I tested it locally also with 200 schools in rsf including assert root hash in the middle.

PR #135, opened 2016-12-14
Allow app to serve multiple registers

philandstuff:
This introduces a new configuration item `otherRegisters`, which
allows you to specify a map from register names to database
configurations for other registers that this app can serve.

The app decides which register to use on a per-request basis, by
inspecting the X-Forwarded-Host or Host headers, and pulling out the
bit before the first dot.  If it matches a named register in
`otherRegisters`, it serves from that register; otherwise, it serves
from itself.

To do this, there is a new class AllTheRegisters, which allows you to
pick out a register using its getRegisterByName() method.  This
returns a RegisterContext instance.
RegisterContext has all the things you need for an individual
register instance - MemoizationStore, DBI, RegisterMetadata, etc.

I also removed RegisterNameConfiguration, as that is no longer a
globally unique thing.  You now need to inject
RegisterReadOnly or RegisterData and get the name from that.

I've deliberately taken an approach such that existing configuration files continue to work, and those existing config files will maintain the current behaviour.  We can then gradually migrate registers into a single pool of instances per environment.


michaelabenyohai:
Feedback discussed IRL. Further steps for a future PR and documented here https://trello.com/c/nUEQ1gdH.

PR #137, opened 2016-12-16
create RegisterName class

philandstuff:
This creates a RegisterName class, which is mostly a thin wrapper
around String for those Strings which specifically name registers.
This achieves a couple of things:

 - a RegisterName can more easily be injected on its own
 - a String can't be passed where a RegisterName is expected, and vice
   versa

This also reduces some duplications: for example, between
AssetsBundleCustomErrorHandler and ThymeleafView (which were both
constructing friendly names) and between
AssetsBundleCustomErrorHandler and RequestContext (which were both
extracting register names from the Host/X-Forwarded-Host header).

### DIFF
--- a/src/main/java/uk/gov/register/AssetsBundleCustomErrorHandler.java
+++ b/src/main/java/uk/gov/register/AssetsBundleCustomErrorHandler.java
@@ -1,7 +1,6 @@
 package uk.gov.register;
 
 import io.dropwizard.setup.Environment;
-import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jetty.server.Request;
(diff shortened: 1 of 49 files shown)


johnollier:
I'm not sure how you would do it in dropwizard, but could you also use something like @Named("register-name") String registerName ?

philandstuff:
@johnollier yes, you can do it that way, and I did consider it.

I think there are other benefits to introducing a `RegisterName` class.  For example, it gives us a place to put the `getFriendlyRegisterName` behaviour.

I think we use String in enough places that we might be accused of having ["Stringly typed"](http://wiki.c2.com/?StringlyTyped) code...

philandstuff:
rebased to fix conflict

PR #140, opened 2016-12-19
Bugfix: stop leaking flyway db connections

philandstuff:
We used to create two separate DataSources per register: one for the
app to use, and a separate one for flyway.  The one created for the
app was created by dbiFactory.build(), which ensures that the
ManagedDataSource instance is correctly registered with the app
lifecycle hooks so that it is shut down when the app shuts down.

The flyway datasource, however, was created manually by us, and we
never register it in the lifecycle hooks.  This means that the app
leaks connections when shut down.  This causes problems when running a
lot of tests, because the collective leaked connections can exhaust
available DB connections and cause test failures.

There are two ways to fix this:

  1. register flyway's datasource with the app lifecycle so it gets
     closed
  2. get flyway to use the same datasource as the rest of the app

I went with option 2 in this commit.  With two separate DataSources,
we have two separate pools, each of which has the same minimum and
maximum sizes taken from the app configuration.  This means that when
you specify `maxSize: 4` in your configuration, you might actually
consume as many as 8 connections.  By using a single shared
DataSource, `maxSize: 4` means that only 4 connections will ever be
used.

### DIFF
--- a/src/main/java/uk/gov/register/RegisterConfiguration.java
+++ b/src/main/java/uk/gov/register/RegisterConfiguration.java
@@ -99,6 +99,7 @@ public class RegisterConfiguration extends Configuration
     private Map<RegisterName, RegisterContextFactory> otherRegisters = new HashMap<>();
 
     public DataSourceFactory getDatabase() {
+        database.getProperties().put("ApplicationName", "openregister");
         return database;
     }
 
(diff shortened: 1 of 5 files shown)


philandstuff:
rebased to fix conflicts

PR #141, opened 2016-12-19
Make trackingId a per-register configuration value

philandstuff:
This rather large PR does several things (sorry):

 - makes `trackingId` a value which is set on a per-register basis.
   This means that you must set the tracking ID for each register in
   the `otherRegisters` config item
 - changes the functional tests to create a single app instance, with
   three different registers. this means that many tests no longer
   need to customize the configuration values any more and can share
   the same RegisterRule instance. In some cases I've been able to
   change @Rules to @ClassRules, cutting test times.
 - refactors the test code so that we make sure we close all Handles
   we create

The new trackingId functionality is tested by the changed
AnalyticsFunctionalTest, which expects different analytical behaviour
from different registers within the same app instance.


src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -97,12 +93,14 @@ public void transactionalRegisterOperation(Consumer<Register> consumer) {
 
     public static void useTransaction(DBI dbi, Consumer<Handle> callback) {
         try {
-            dbi.useTransaction(TransactionIsolationLevel.SERIALIZABLE, (handle, status) -> {
-                handle.getConnection().setClientInfo("ApplicationName", "openregister_transactional");
```
chr1sdi: Naming the client no likey? Was it useful over there from the start? What was/is the benefit if any that we may lose like generating a `random number` + `register name` and then setting it here and log it out in logs as well?

src/test/java/uk/gov/register/db/mappers/EntryMapperTest.java
```diff
@@ -15,22 +14,21 @@
...
-        DBI dbi = new DBI("jdbc:postgresql://localhost:5432/ft_openregister_java?user=postgres&ApplicationName=EntryMapperTest");
-        Handle h = dbi.open();
-        EntryQueryDAO entryQueryDAO = dbi.onDemand(EntryQueryDAO.class);
+        DBI dbi = new DBI("jdbc:postgresql://localhost:5432/ft_openregister_java_address?user=postgres&ApplicationName=EntryMapperTest");
 
-        h.execute("insert into entry(entry_number, sha256hex, timestamp) values(5, 'abcdef', :timestamp)",expectedInstant.getEpochSecond());
-
-        // this method implicitly invokes EntryMapper
-        Collection<Entry> allEntriesNoPagination = entryQueryDAO.getAllEntriesNoPagination();
+        Collection<Entry> allEntriesNoPagination = dbi.withHandle(h -> {
```
chr1sdi: The only thing here is that it all will be within the same transaction, previously the changes were committed before querying - just an observation really :)

src/main/java/uk/gov/register/core/RegisterContext.java
philandstuff: a) I didn't think it was useful, no
b) I don't know whether it permanently renames the connection (so it keeps the `_transactional` name even after returning to the pool) which would do more harm than good

src/main/java/uk/gov/register/core/RegisterContextFactory.java
```diff
@@ -13,19 +13,25 @@
...
+    private String trackingId;
+
     @SuppressWarnings("unused, used by Jackson")
     public RegisterContextFactory() {
     }
 
-    public RegisterContextFactory(DataSourceFactory database) {
+    public RegisterContextFactory(DataSourceFactory database, Optional<String> trackingId) {
         this.database = database;
+        this.trackingId = trackingId.orElse(null);
```
philandstuff: I couldn't get this to work with Jackson, sorry. If you know how, happy to make the change

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
```diff
@@ -38,56 +36,65 @@
...
+    public AnalyticsFunctionalTest(String targetUrl) {
         this.targetUrl = targetUrl;
-        this.shouldIncludeAnalytics = shouldIncludeAnalytics;
     }
 
-    @Parameterized.Parameters(name = "{index} - url: {0} with code: {1} should include analytics: {2}")
+    @Parameterized.Parameters(name = "{index} - url: {0}")
     public static List data() {
-        List sourceData = new ArrayList();
+        List<Object[]> sourceData = new ArrayList<>();
```
chr1sdi: We can simplify this to 
```
return Arrays.asList(
                "/",
                "/download",
bla bla bla
```

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
```diff
@@ -96,24 +103,28 @@ public void checkAnalyticsScriptsPresence() throws Exception {
...
+        Boolean docIncludesAnalyticsId = doc.getElementById("analytics-tracking-id") != null;
+        Boolean docIncludesMainAnalytics = doc.getElementById("analytics-main") != null;
+        Boolean docIncludesExtLinksAnalytics = doc.getElementById("analytics-external-links") != null;
+
+        assertThat(docIncludesAnalyticsId, equalTo(true));
+        assertThat(docIncludesMainAnalytics, equalTo(true));
+        assertThat(docIncludesExtLinksAnalytics, equalTo(true));
     }
 
+    private static Object[] generateTestSetFor(String url) {
```
chr1sdi: Then this can go away as well :)

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
philandstuff: nope, it still needs to be a `List<Object[]>` but your syntax gives a `List<String>`

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
```diff
@@ -38,56 +36,65 @@
...
-                testEntry2.itemJson);
+        register.mintLines("address", testEntry1.itemJson, testEntry2.itemJson);
     }
 
     private final String targetUrl;
-    private final Boolean shouldIncludeAnalytics;
 
-    public AnalyticsFunctionalTest(String targetUrl, String trackingId, Boolean shouldIncludeAnalytics) {
-        this.register = createRegister(trackingId);
+    public AnalyticsFunctionalTest(String targetUrl) {
```
philandstuff: I'll extract the private method.

You say "not clearly visible in the test" but I created the constants `REGISTER_WITH_MISSING_TRACKING_ID` etc specifically to document these assumptions. Is that not enough? Can you suggest anything more?

src/test/java/uk/gov/register/functional/ApplicationTest.java
```diff
@@ -30,7 +30,7 @@ public void setup() {
     @Test
     public void appSupportsCORS() {
         String origin = "http://originfortest.com";
-        Response response = register.target().path("/entries")
+        Response response = register.targetRegister("address").path("/entries")
```
chr1sdi: This is the same as `getRequest()` do think we should keep only one of them? maybe possibly or make a convention when to use what?

src/test/java/uk/gov/register/functional/ApplicationTest.java
```diff
@@ -20,7 +20,7 @@
 
 public class ApplicationTest {
     @ClassRule
-    public static RegisterRule register = new RegisterRule("address");
+    public static RegisterRule register = new RegisterRule();
```
chr1sdi: Would it be possible to pass in `registerName` as 1 of the constructors to then skip passing register name in `getRequest` or `targetRegister` overloads (just a sugah really).

philandstuff: no. the whole point of removing the register name is to make the `RegisterRule` more sharable between different tests, to allow us to move from `@Rule` to `@ClassRule` (and from there on to `@Suite` which might allow us to only start a single app server, massively cutting test times. This is as recommended in the dropwizard-testing docs.

Every parameter we add to `RegisterRule` reduces this scope for sharing resources between functional tests and cutting test times.

src/test/java/uk/gov/register/functional/app/RegisterRule.java
```diff
@@ -17,88 +21,121 @@
...
-        ConfigOverride[] configOverrides = constructOverrides(register, overrides);
+    private List<Handle> handles = new ArrayList<>();
+
+    public RegisterRule(ConfigOverride... overrides) {
         this.appRule = new DropwizardAppRule<>(RegisterApplication.class,
                 ResourceHelpers.resourceFilePath("test-app-config.yaml"),
-                configOverrides);
-        wipeRule = new WipeDatabaseRule();
+                overrides);
+        wipeRule = new WipeDatabaseRule("address", "register");
```
philandstuff: oops, I missed that

src/test/java/uk/gov/register/functional/AssetsBundleCustomErrorHandlerTest.java
```diff
@@ -16,11 +16,11 @@
 
 public class AssetsBundleCustomErrorHandlerTest {
     @ClassRule
-    public static RegisterRule register = new RegisterRule("address");
+    public static RegisterRule register = new RegisterRule();
```
chr1sdi: Can this be changed to `@Rule` without the static for consistency?

src/test/java/uk/gov/register/functional/DataDownloadFunctionalTest.java
```diff
@@ -29,7 +29,7 @@
 public class DataDownloadFunctionalTest {
 
     @ClassRule
-    public static RegisterRule register = new RegisterRule("address");
+    public static RegisterRule register = new RegisterRule();
```
chr1sdi: Kapow same as above - a possible change to`@Rule` without the static for consistency?

src/test/java/uk/gov/register/functional/AssetsBundleCustomErrorHandlerTest.java
philandstuff: consistency with what?

src/test/java/uk/gov/register/functional/DataDownloadFunctionalTest.java
philandstuff: ditto.

more generally, I have a goal that `RegisterRule` should only ever be in a `@ClassRule` - and that's the consistency I'm aiming for

src/test/java/uk/gov/register/functional/ApplicationTest.java
philandstuff: `getRequest()` is a convenience wrapper around `targetRegister()` for GET requests.

I'm in two minds. I think `getRequest` reads nicer but it doesn't do very much. /shrug

src/test/java/uk/gov/register/functional/DeleteRegisterDataFunctionalTest.java
```diff
@@ -6,34 +6,36 @@
...
 
+        Response deleteResponse = register.deleteRegisterData("register");
         assertThat(deleteResponse.getStatus(), equalTo(200));
-        assertThat(testItemDAO.getItems(), is(empty()));
-        assertThat(testEntryDAO.getAllEntries(), is(empty()));
+
+        Response entriesResponse2 = register.getRequest("register", "/entries.json");
+        String entriesRawJSON = entriesResponse2.readEntity(String.class);
+
+        assertThat(entriesRawJSON, is("[]"));
```
chr1sdi: tru dat as in this is much better

src/test/java/uk/gov/register/functional/EntriesResourceFunctionalTest.java
```diff
@@ -25,30 +26,32 @@
...
 public class EntriesResourceFunctionalTest {
     @ClassRule
-    public static RegisterRule register = new RegisterRule("address");
+    public static RegisterRule register = new RegisterRule();
+    private WebTarget addressTarget;
 
     @Before
     public void setup() {
         register.wipe();
+        addressTarget = register.targetRegister("address");
```
chr1sdi: Is this required to be run before each test? Maybe just maybe we can make it nicely into:
`private final WebTarget addressTarget = register.targetRegister("address");`
what do you think? It does have `final` on it too ;)

src/test/java/uk/gov/register/db/mappers/EntryMapperTest.java
philandstuff: actually, it doesn't. `dbi.withHandle` just makes a `Handle` (which wraps a `Connection`) available.  It doesn't imply any sort of transactional operation - that needs `dbi.{in,use}Transaction`.

(I had to check this locally before replying though)

src/test/java/uk/gov/register/functional/EntryResourceFunctionalTest.java
```diff
@@ -28,21 +29,23 @@
...
     private final String item2Hash = "sha-256:" + DigestUtils.sha256Hex(item2);
-    
+    private WebTarget addressTarget;
+
     @Before
     public void publishTestMessages() throws Throwable {
         register.wipe();
-        register.mintLines(item1, item2);
+        register.mintLines("address", item1, item2);
+        addressTarget = register.targetRegister("address");
```
chr1sdi: If the previous comment in `EntriesResourceFunctionalTest.java` didn't infuriate you maybe we can do the same for this class?

src/test/java/uk/gov/register/functional/EntriesResourceFunctionalTest.java
philandstuff: huh, I thought that wouldn't work, but it does! will change :)

src/test/java/uk/gov/register/functional/HomePageFunctionalTest.java
```diff
@@ -13,22 +13,22 @@
 import static org.hamcrest.Matchers.greaterThan;
 import static org.junit.Assert.assertThat;
 
-public class HomePageFunctionalTest  {
+public class HomePageFunctionalTest {
 
     private final String registerName = "postcode";
```
chr1sdi: Other test don't have `registerName` as a variable... just saying ;)

src/test/java/uk/gov/register/functional/EntryResourceFunctionalTest.java
philandstuff: 👍

src/test/java/uk/gov/register/functional/HomePageFunctionalTest.java
philandstuff: it was like that when I got here, honest! :)

src/test/java/uk/gov/register/db/mappers/EntryMapperTest.java
chr1sdi: now since you know the 'ok man' I shall use it in this pull request...
.
.
.
:ok_man:

src/test/java/uk/gov/register/functional/ApplicationTest.java
chr1sdi: :ok_man:

src/test/java/uk/gov/register/functional/ApplicationTest.java
chr1sdi: :ok_man:

src/test/java/uk/gov/register/functional/AssetsBundleCustomErrorHandlerTest.java
chr1sdi: :ok_man: as I read the next comment

src/test/java/uk/gov/register/functional/DataDownloadFunctionalTest.java
chr1sdi: :ok_man: so can we change `@Rule` to `@ClassRule` in: `DeleteRegisterDataFunctionalTest`?

src/test/java/uk/gov/register/functional/LoadSerializedFunctionalTest.java
```diff
@@ -26,17 +25,26 @@
...
 public class LoadSerializedFunctionalTest {
-    public static final int APPLICATION_PORT = 9000;
     private static final String registerName = "register";
 
     @Rule
     public TestRule wipe = new WipeDatabaseRule();
 
     @ClassRule
-    public static final RegisterRule register = new RegisterRule(registerName);
+    public static final RegisterRule register = new RegisterRule();
```
chr1sdi: Do you think it may be worth extracting a base class at this point or not yet because of a reason? Somink like this or is it too stinky and they accidentally look the same? Just a food for thought ;) no hate just love...
```
public abstract class BaseTest {
    @ClassRule
    public static final RegisterRule register = new RegisterRule();

    protected final CanonicalJsonMapper canonicalJsonMapper = new CanonicalJsonMapper();
    protected static TestRecordDAO testRecordDAO;
    protected static TestEntryDAO testEntryDAO;
    protected static TestItemCommandDAO testItemDAO;

    @BeforeClass
    public static void setUpClass() throws Exception {
        Handle handle = register.handleFor("register");
        testRecordDAO = handle.attach(TestRecordDAO.class);
        testEntryDAO = handle.attach(TestEntryDAO.class);
        testItemDAO = handle.attach(TestItemCommandDAO.class);
    }
}
```

src/test/java/uk/gov/register/functional/HomePageFunctionalTest.java
chr1sdi: :ok_man:

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
```diff
@@ -38,56 +36,63 @@
...
     }
 
     @Test
-    public void checkAnalyticsScriptsPresence() throws Exception {
-        Response response = register.getRequest(targetUrl, TEXT_HTML);
+    public void validTrackingId_shouldIncludeAnalyticsCode() throws Exception {
+        assertUrlHasResponseWithAppropriateAnalytics(REGISTER_WITH_VALID_TRACKING_ID, true);
+    }
+
+    public void assertUrlHasResponseWithAppropriateAnalytics(String registerName, boolean shouldIncludeAnalytics) {
```
chr1sdi: Let's keep our secrets secret heh can we change this to `private` please?

src/test/java/uk/gov/register/functional/FindEntityTest.java
```diff
@@ -18,17 +18,17 @@
...
     public void publishTestMessages() {
         register.wipe();
-        register.mintLines("{\"street\":\"ellis\",\"address\":\"12345\"}", "{\"street\":\"presley\",\"address\":\"6789\"}", "{\"street\":\"ellis\",\"address\":\"145678\"}");
+        register.mintLines("address", "{\"street\":\"ellis\",\"address\":\"12345\"}", "{\"street\":\"presley\",\"address\":\"6789\"}", "{\"street\":\"ellis\",\"address\":\"145678\"}");
     }
 
     @Test
     public void find_shouldReturnEntryWithThPrimaryKey_whenSearchForPrimaryKey() throws JSONException {
-        WebTarget target = register.target();
+        WebTarget target = register.targetRegister("address");
```
chr1sdi: missing 'e' in the test name.... I know it's not you but I'll let you have a joy of finding it :)

src/test/java/uk/gov/register/functional/RecordListResourceFunctionalTest.java
```diff
@@ -24,18 +24,20 @@
...
-    public static RegisterRule register = new RegisterRule("address");
+    public static RegisterRule register = new RegisterRule();
+    private WebTarget addressTarget;
 
     @Before
     public void publishTestMessages() {
         register.wipe();
-        register.mintLines("{\"street\":\"ellis\",\"address\":\"12345\"}", "{\"street\":\"presley\",\"address\":\"6789\"}", "{\"street\":\"ellis\",\"address\":\"145678\"}", "{\"street\":\"updatedEllisName\",\"address\":\"145678\"}", "{\"street\":\"ellis\",\"address\":\"6789\"}");
+        register.mintLines("address", "{\"street\":\"ellis\",\"address\":\"12345\"}", "{\"street\":\"presley\",\"address\":\"6789\"}", "{\"street\":\"ellis\",\"address\":\"145678\"}", "{\"street\":\"updatedEllisName\",\"address\":\"145678\"}", "{\"street\":\"ellis\",\"address\":\"6789\"}");
+        addressTarget = register.targetRegister("address");
```
chr1sdi: Can this be a nice `final` as well?

src/test/java/uk/gov/register/functional/app/RegisterRule.java
```diff
@@ -17,88 +22,122 @@
...
+        return authenticatedTarget(registerName).path("/delete-register-data").request().delete();
     }
 
     public void wipe() {
         wipeRule.before();
     }
 
     private JerseyClientBuilder clientBuilder() {
+        InMemoryDnsResolver customDnsResolver = new InMemoryDnsResolver();
+        for (String registerHostname : registerHostnames.values()) {
```
chr1sdi: a POJF (Plain Old Java Forloop) ;)

src/test/java/uk/gov/register/functional/db/EntryIteratorDAOFunctionalTest.java
```diff
@@ -16,33 +13,31 @@
...
 
     @ClassRule
-    public static final RegisterRule register = new RegisterRule("register");
+    public static final RegisterRule register = new RegisterRule();
+    private TestEntryDAO testEntryDAO;
 
     @Before
     public void publishTestMessages() {
         register.wipe();
+        Handle handle = register.handleFor("register");
```
chr1sdi: Do these have to be set for each test?

src/test/java/uk/gov/register/functional/LoadSerializedFunctionalTest.java
philandstuff: can they not just be extra methods on the RegisterRule, or a separate helper class? (I prefer composition over inheritance :P )

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
philandstuff: 👏 will fix

src/test/java/uk/gov/register/functional/FindEntityTest.java
philandstuff: ouch

src/main/java/uk/gov/register/core/RegisterContextFactory.java
chr1sdi: How about this?
https://github.com/FasterXML/jackson-datatype-jdk8
`jackson_version = '2.8.3'` from `gradle.build` 😞

src/test/java/uk/gov/register/functional/LoadSerializedFunctionalTest.java
chr1sdi: fair enough :ok_man:

src/test/java/uk/gov/register/functional/AnalyticsFunctionalTest.java
chr1sdi: :ok_man:

src/test/java/uk/gov/register/functional/FindEntityTest.java
chr1sdi: I knooow right?
:ok_man:

chr1sdi:
:ok_man: 👍

philandstuff:
@chr1sdi rebased

PR #142, opened 2016-12-19
Reload configs on data delete

karlbaker02:
This is a refactor of a bunch of places where we currently use RegistersConfiguration and FieldsConfiguration to instead reference the ConfigManager, which makes it easy to update configs at runtime and update the corresponding in-memory stores corresponding to these (rather than attempting to modify what's returned from the DI layer).


src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -17,22 +16,19 @@
...
+    public RegisterContext(RegisterName registerName, ConfigManager configManager, MemoizationStore memoizationStore, DBI dbi, Flyway flyway) {
         this.registerName = registerName;
-        this.registersConfiguration = registersConfiguration;
-        this.fieldsConfiguration = fieldsConfiguration;
+        this.configManager = configManager;
         this.memoizationStore = memoizationStore;
         this.dbi = dbi;
         this.flyway = flyway;
-        this.registerMetadata = registersConfiguration.getRegisterMetadata(registerName);
+        this.registerMetadata = configManager.getRegistersConfiguration().getRegisterMetadata(registerName);
```
philandstuff: I wonder if (in future) we could move the `getRegisterMetadata` method on to `ConfigManager`?

src/main/java/uk/gov/register/service/ItemConverter.java
```diff
@@ -14,17 +14,17 @@
...
-        this.fieldsConfiguration = fieldsConfiguration;
+    public ItemConverter(ConfigManager configManager) {
+        this.configManager = configManager;
     }
 
     public FieldValue convert(Map.Entry<String, JsonNode> mapEntry) {
         String fieldName = mapEntry.getKey();
         JsonNode value = mapEntry.getValue();
-        FieldConverter fieldConverter = new FieldConverter(fieldsConfiguration.getField(fieldName));
+        FieldConverter fieldConverter = new FieldConverter(configManager.getFieldsConfiguration().getField(fieldName));
```
philandstuff: ditto `getField()`

PR #144, opened 2016-12-20
Add functional tests for entry, item and record

karlbaker02:
Adding tests to ensure that entry, item and record pages return a 200 status code when we expect the thing to exist.


philandstuff:
:+1:

PR #145, opened 2016-12-20
More register-specific config items

philandstuff:
This makes the register download and delete endpoints into register-specific config values.


karlbaker02:
All looks good 👍

PR #149, opened 2016-12-21
Demonstrate delete data problem

karlbaker02:
A few notes about this pull request:
- In `RegisterMemoizationStoreFactory` it would be great if we didn't have to specify the types of `MemoizationStore` that we care about - is it overkill to add a new service which would take these types when the application is registered and loop through this in the factory to create a new store?
- Assumption has been made that we will always be returning an `InMemory` store if the requested type isn't found, but because the factory specifies `<T extends MemoizationStore>` then this should never be the case.

### DIFF
--- a/src/main/java/uk/gov/register/core/RegisterContext.java
+++ b/src/main/java/uk/gov/register/core/RegisterContext.java
@@ -6,16 +6,16 @@ import org.skife.jdbi.v2.DBI;
 import org.skife.jdbi.v2.Handle;
 import org.skife.jdbi.v2.TransactionIsolationLevel;
 import org.skife.jdbi.v2.exceptions.CallbackFailedException;
-import uk.gov.register.configuration.ConfigManager;
-import uk.gov.register.configuration.DeleteRegisterDataConfiguration;
-import uk.gov.register.configuration.RegisterFieldsConfiguration;
-import uk.gov.register.configuration.RegisterTrackingConfiguration;
-import uk.gov.register.configuration.ResourceConfiguration;
+import uk.gov.register.configuration.*;
 import uk.gov.register.db.*;
+import uk.gov.register.exceptions.NoSuchConfigException;
 import uk.gov.register.service.ItemValidator;
+import uk.gov.verifiablelog.store.memoization.InMemoryPowOfTwoNoLeaves;
 import uk.gov.verifiablelog.store.memoization.MemoizationStore;
 
+import java.io.IOException;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 public class RegisterContext implements
@@ -24,7 +24,7 @@ public class RegisterContext implements
         ResourceConfiguration {
     private RegisterName registerName;
     private ConfigManager configManager;
-    private MemoizationStore memoizationStore;
+    private AtomicReference<MemoizationStore> memoizationStore;
     private DBI dbi;
     private Flyway flyway;
     private final Optional<String> trackingId;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/register/core/RegisterMemoizationStoreFactory.java
```diff
@@ -0,0 +1,22 @@
...
+
+import uk.gov.verifiablelog.store.memoization.InMemory;
+import uk.gov.verifiablelog.store.memoization.InMemoryPowOfTwoNoLeaves;
+import uk.gov.verifiablelog.store.memoization.MemoizationStore;
+
+public class RegisterMemoizationStoreFactory {
+    private String memoizationStoreTypeName;
+
+    public <T extends MemoizationStore> RegisterMemoizationStoreFactory(Class<T> memoizationStoreClass) {
+        this.memoizationStoreTypeName = memoizationStoreClass.getSimpleName();
```
philandstuff: I think it's good to think about how we might configure the specific MemoizationStore that we use, rather than hardcoding, but I don't think we should do that as part of fixing a bug.

Let's have a conversation about this.

src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -62,7 +61,7 @@ public Flyway getFlyway() {
         return flyway;
     }
 
-    public Register buildOnDemandRegister() {
+    public synchronized Register buildOnDemandRegister() {
```
philandstuff: I'm concerned about `synchronized` here, and I think we might be able achieve what we want using something like `AtomicReference`s instead. In any case, where we have concurrency concerns (which we certainly do here), we should think about how we document those concerns so that it's clear why the various concurrency primitives are present and how to avoid introducing subtle bugs when maintaining this code.

src/main/java/uk/gov/register/resources/DeleteRegisterDataResource.java
```diff
@@ -30,6 +32,7 @@ public DeleteRegisterDataResource(RegisterContext registerContext, ConfigManager
     public Response deleteRegisterData() throws NoSuchConfigException, IOException {
         flyway.clean();
         configManager.refreshConfig();
+        registerContext.resetMemoizationStore();
         flyway.migrate();
```
philandstuff: I feel like this behaviour should be encapsulated into a `resetRegister()` method on `RegisterContext`? It's certainly not the resource's job to know that we need to reset both the db and the memoization store together.

src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -98,6 +98,15 @@ public void transactionalRegisterOperation(Consumer<Register> consumer) {
         transactionalMemoizationStore.commitHashesToStore();
     }
 
+    public void resetRegister() throws IOException, NoSuchConfigException {
+        if (enableRegisterDataDelete) {
```
philandstuff: should we have a test for this conditional?

src/main/java/uk/gov/register/resources/DeleteRegisterDataResource.java
```diff
@@ -14,23 +12,19 @@
...
-    private Flyway flyway;
-    private ConfigManager configManager;
+    private RegisterContext registerContext;
 
     @Inject
-    public DeleteRegisterDataResource(RegisterContext registerContext, ConfigManager configManager) {
-        this.flyway = registerContext.getFlyway();
-        this.configManager = configManager;
+    public DeleteRegisterDataResource(RegisterContext registerContext) {
+        this.registerContext = registerContext;
```
philandstuff: this resource is really nice now, super lean and mean 💪

PR #151, opened 2016-12-21
Update frontend toolkit

danmitchell-:
- Remove old frontend toolkit and template from repo that were included via subtree
- Add latest frontend toolkit and template as submodule dependancies
- Refactor CSS to fix compiling


philandstuff:
🎉

PR #152, opened 2016-12-21
Introduce test register enum

philandstuff:
This introduces an enum `TestRegister`, to capture configuration information about the registers defined in `test-app-config.yaml` and for use with `RegisterRule`.  It means we can pass in an enum value to `getRequest()` and similar methods, rather than passing in a String representing a register.

This will also be useful for my next PR, where I intend to make different register have different basic auth credentials, and TestRegister will be able to hold that data.


karlbaker02:
👍

PR #150, opened 2016-12-21
make sure all white space is 2 spaces

danmitchell-:
Clean up frontend code based on code styleguides.

No code changes. 💯

PR #154, opened 2016-12-21
Add wrapper to all tables with scroll property

danmitchell-:
Add a parent div tag with overflow scroll to allow horizontal scroll

### DIFF
--- a/src/main/resources/assets/sass/main.scss
+++ b/src/main/resources/assets/sass/main.scss
@@ -163,12 +163,14 @@ main {
 $table-header: #f5f5f5;
 $table-row: #f9f9f9;
 
+.table-wrapper {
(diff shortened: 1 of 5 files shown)


src/main/resources/assets/sass/main.scss
```diff
@@ -163,12 +163,14 @@ main {
 $table-header: #f5f5f5;
 $table-row: #f9f9f9;
 
+.table {
```
tvararu: I think `.table-wrapper` is a bit better, what do you think?

tvararu:
I assume the whitespace stuff is dependant on https://github.com/openregister/openregister-java/pull/150?

danmitchell-:
I've closed #150 and will re-do after this.

PR #155, opened 2016-12-21
remove last use of ConfigOverride

philandstuff:
This was dead code -- the DeleteRegisterDataFunctionalTest didn't need
the ConfigOverride at all.  We can now get rid of the
ConfigOverride... constructor from RegisterRule.


karlbaker02:
👍

PR #157, opened 2016-12-22
Move table spacing to wrapper

danmitchell-:
Moving the margin to the wrapper to remove the margin above the scroll
bar

## Before
screen shot 2016-12-22 at 11 14 19

## After
screen shot 2016-12-22 at 11 15 52


karlbaker02:
Looks good 👍

PR #158, opened 2016-12-22
fix register register template error

philandstuff:
This was causing a type conversion error (between String and RegisterName)


karlbaker02:
👍

PR #159, opened 2016-12-22
make historyPageUrl a register-specific config item

philandstuff:
Tested locally. The current config.yaml now means that the local
school register has a history page url, but the local register
register does not.


karlbaker02:
👍  Tested locally too, all good.

PR #160, opened 2016-12-22
test independence of MemoizationStores

philandstuff:
This adds a test to verify that we don't accidentally share a
MemoizationStore between two different registers -- ie, when we get
proofs from one register, its hashes should be completely different
from the proofs in a different register.


karlbaker02:
Looks good 👍

PR #161, opened 2016-12-22
Remove whitespace

danmitchell-:


michaelabenyohai:
Please can you rebase this as it now has merge conflicts.

src/main/resources/templates/home.html
```diff
@@ -3,84 +3,118 @@
...
-        </div>
+      <summary>View register fingerprint</summary>
+      <div class="fingerprint-container">
+        <span class="fingerprint-explanation">
+          <a href="https://gdstechnology.blog.gov.uk/2015/10/13/guaranteeing-the-integrity-of-a-register/">
+            See how the fingerprint works
+          </a>
+        </span>
+        <pre class="fingerprint">
+          <code>
```
michaelabenyohai: Unfortunately this change has caused an unexpected change in the "register fingerpint" section. This appears to be an annoying way that Tymeleaf works, where whitespace makes a difference to appearance. It might be worth changing this back to how it was - I know it is a bit of a hack, but seeing as we want to build a new front end at some point, I wouldn't want to waste too much time fixing this.

Now looks like this:

screen shot 2016-12-23 at 12 18 33


Should look like this:
screen shot 2016-12-23 at 12 18 50

PR #162, opened 2016-12-22
Check basic auth credentials for the requested register only

philandstuff:
This makes basic auth credentials a per-register setting.  You can
specify completely independent usernames and passwords for each
register.  Only the credentials for the requested register will be
checked.

This also makes credentials a *required* setting: it is no longer
permissible not to specify basic auth credentials in config.

In order to check credentials only for the requested register, we had
to be able to inject a `RegisterContext` into the `AuthFilter`, which
means that the dependency injection container needs to build the
`AuthFilter` for itself.  However dropwizard-auth's AuthDynamicFeature
doesn't allow this, because it requires an *instance* of a
ContainerRequestFilter, not merely a class.

Therefore, we created RegisterAuthDynamicFeature, which is just a
copypaste of AuthDynamicFeature but which allows registering a class.
We use this to register BasicAuthFilter, which takes an injected
`Provider<RegisterContext>`, and uses that to get the correct
credentials to check against.

Paired with @karlbaker02.


michaelabenyohai:
@karlbaker02 please can you take a look at the failing test?

karlbaker02:
Will look at the failing tests @michaelabenyohai

PR #165, opened 2016-12-28
Fix bug where we leak connections from EntryIteratorDAO

michaelabenyohai:
Close ResultIterator when an entire VerifiableLog action has been completed

We now use JDBI onDemand SQL objects which manage their own opening and closing of
database handles. When we reading from the database via EntryIteratorDAO, if the iterator
is not completely consumed (until there is no more hasNext), the iterator does not get closed
and therefore neither does the database handle. This therefore leaks connections.

We did not see this previously because we were explicitely closing the database handle
by using dbi.inTransaction around each action on the verifiableLog.


chr1sdi:
👍

PR #166, opened 2016-12-29
Make RSF download streaming work

chr1sdi:
Fix RSF download streaming which works like a charm.

First problem was in buffering `OutputStream` and unknown content length hence it was buffering the max length. Solution is to set `ServerProperties.OUTBOUND_CONTENT_LENGTH_BUFFER` to `false`.
http://stackoverflow.com/questions/29637151/jersey-streamingoutput-as-response-entity#answer-34358215

The other issue was that in order to create a JDBI Postgres cursor query we need to set `autoCommit` to `false`.
https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor


chr1sdi:
'does not compute' at the moment, closing for now

PR #168, opened 2016-12-29
Fix bug where we accidentally use zero-indexed entry-numbers for entry proofs

michaelabenyohai:
This meant that we were calculating incorrect entry proofs. In the app entry-numbers are one-indexed, but the VerifiableLog library assumes Merkle Tree leaves are zero-indexed. The bug existed because we were not correctly mapping between the two indexed.

It resulted in /proof/entry/1/2/merkle:sha-256 and /proof/entry/2/2/merkle:sha-256 returning the same result.

There was also a problem where we allow /proof/entry/0/2/merkle:sha-256 and /proof/entry/3/2/merkle:sha-256 to be called and they return results (which it shouldn't because these requests do not make sense). We now return a BadRequestException for parameter combinations like this.


chr1sdi:
👍

PR #169, opened 2016-12-30
Add standardjs badge to readme

danmitchell-:
We use StandardJS to lint our JS so adding this to make it more obvious

screen shot 2016-12-30 at 10 30 31


chr1sdi:
👍

PR #170, opened 2017-01-03
Allow unknown properties to exist in config file and remove some obsolete constructors

karlbaker02:


src/main/java/uk/gov/register/configuration/FieldsConfiguration.java
```diff
@@ -21,13 +21,6 @@ public FieldsConfiguration(Optional<String> fieldsResourceYamlPath) {
                 });
     }
 
-    public FieldsConfiguration(byte[] fieldsConfig) {
```
karlbaker02: On  the theme of 'Clean as you go', this constructor isn't used.

src/main/java/uk/gov/register/configuration/RegistersConfiguration.java
```diff
@@ -23,14 +23,6 @@ public RegistersConfiguration(Optional<String> registersResourceYamlPath) {
         );
     }
 
-    public RegistersConfiguration(byte[] registersConfig) {
```
karlbaker02: On  the theme of 'Clean as you go', this constructor isn't used.

PR #173, opened 2017-01-04
Redesign homepage and standardise page structure

danmitchell-:
Whilst redesigning the homepage there was a few inconsistencies with
the page structure throughout the app so I updated throughout,
including changes from the gov_toolkit, gov_elements and gov_components.

## Before
screen shot 2017-01-05 at 09 48 01

## After
screen shot 2017-01-09 at 10 46 14


tvararu:
Code looks good to me. 👍 

Can you include two screenshots, before and after, of how the page looks with the new `<div id="wrapper">`?

danmitchell-:
@tvararu the `#wrapper` addition was just to match the gov.uk structure was just to match gov.uk

src/main/resources/templates/home.html
```diff
@@ -4,113 +4,48 @@
...
-          and
-          <a href="https://register.register.gov.uk">view all registers</a>.
+          <a href="/records" class="button">View Register</a>
         </p>
+        <p class="text">You can also <a href="/download">download a copy</a> of the register.</p>
+        <p class="text">The <span th:utext="${friendlyRegisterName}"></span> contains <span th:text="${totalRecords}"></span> records and includes the following fields:</p>
+        <ul>
+          <li th:each="field : ${register.fields}" th:text="${field}"></li>
+        </ul>
+        <h3 class="heading-small">Using this regsiter</h3>
```
michaelabenyohai: `regsiter` should be `register`

src/main/resources/templates/home.html
```diff
@@ -4,113 +4,48 @@
...
+        <p>Download a copy of this register. You'll need to reguallry uodate it to make sure you have the most current verion of the data.</p>
```
michaelabenyohai: `reguallry uodate` should be `regularly update` and `verion` should be `version`

src/main/resources/templates/home.html
```diff
@@ -4,113 +4,48 @@
...
+
+      <div class="column-third">
+        <aside class="govuk-related-items" role="complementary">
+          <h2 class="heading-medium">About this register</h2>
+          <th:block th:replace="main.html::attribution"></th:block>
+          <h3 class="heading-small">Last updated:</h3>
+          <p th:text="${lastUpdatedTime}">Last updated</p>
+          <h3 class="heading-small">Elsewhere on GOV.UK</h3>
+          <p>
+            <a th:href="${contentPages.registersIntroductionUrl}" target="_blank">Introducting registers</a>
```
michaelabenyohai: `Introducting` should be `Introducing`

src/main/resources/templates/home.html
```diff
@@ -4,113 +4,48 @@
...
+        <h3 class="heading-small">Using this regsiter</h3>
+        <p>Download a copy of this register. You'll need to reguallry uodate it to make sure you have the most current verion of the data.</p>
+        <p>If you're building a service, you can use this data to create elements for forms.</p>
+        <p>You might need to combine data from this register with data from another source to do this.</p>
+      </main>
+
+      <div class="column-third">
+        <aside class="govuk-related-items" role="complementary">
+          <h2 class="heading-medium">About this register</h2>
+          <th:block th:replace="main.html::attribution"></th:block>
```
michaelabenyohai: This is nice except that in the design the `Managed by` heading has the same heading-size as `Last updated`. Here it the first is a `heading-medium` and the second is a `heading-small`. I understand it's because we use `heading-medium` for all other pages but is there away we can achieve the correct size on both?

src/main/resources/templates/home.html
```diff
@@ -4,113 +4,48 @@
...
+          <h3 class="heading-small">Last updated:</h3>
+          <p th:text="${lastUpdatedTime}">Last updated</p>
+          <h3 class="heading-small">Elsewhere on GOV.UK</h3>
```
michaelabenyohai: Looks like in the design that `Elsewhere on GOV.UK` should be the same size as `About this register` so should this be a `heading-medium`?

src/main/resources/templates/home.html
michaelabenyohai: @danmitchell- it would be great if we can get these heading sizes correct before merging. It looks like it should be:

"About this register" - heading-medium
"Managed by" - heading-small
"Last-updated" - heading-small
"Elsewhere on GOV.UK" - heading-medium

src/main/resources/templates/home.html
```diff
@@ -4,113 +4,46 @@
...
-          <a href="https://register.register.gov.uk">view all registers</a>.
+          <a href="/records" class="button">View Register</a>
         </p>
+        <p class="text">You can also <a href="/download">download a copy</a> of the register.</p>
+        <p class="text">The <span th:utext="${friendlyRegisterName}"></span> contains <span th:text="${totalRecords}"></span> records and includes the following fields:</p>
+        <ul class="list-bullet">
+          <li th:each="field : ${register.fields}" th:text="${field}"></li>
+        </ul>
+        <h3 class="heading-small">Using this register</h3>
+        <p>Download a copy of this register. You'll need to regularly uodate it to make sure you have the most current version of the data.</p>
```
michaelabenyohai: `uodate` should be `update`

michaelabenyohai:
These are the things that would be good to fix before this is merged:

- heading sizes of the things on the right-hand side (mentioned in another comment).
- can you remove GOV.UK from the banner altogether? I believe we shouldn't be using this for now.
- it looks like the "School register" heading on the left (not the one in the banner) should be level with the blue line on the right (is this easy to fix?).
- one small typo as mentioned in another comment.
- add other external links (waiting to be told by design as to where these should go).

PR #174, opened 2017-01-04
Fix RSF download stream

chr1sdi:
part of: https://trello.com/c/q8anQW60/542-make-download-load-rsf-stream


src/main/java/uk/gov/register/db/EntryQueryDAO.java
```diff
@@ -31,25 +31,25 @@
...
     @RegisterMapper(EntryMapper.class)
-    @SqlQuery("select * from entry where entry_number >= :start and entry_number \\< :start + :limit order by entry_number asc")
+    @SqlQuery("select entry_number, sha256hex, timestamp, key from entry where entry_number >= :start and entry_number \\< :start + :limit order by entry_number asc")
     Collection<Entry> getEntries(@Bind("start") int start, @Bind("limit") int limit);
 
-    @SqlQuery("SELECT * FROM entry WHERE entry_number >= :entryNumber ORDER BY entry_number")
+    @SqlQuery("SELECT entry_number, sha256hex, timestamp, key FROM entry WHERE entry_number >= :entryNumber ORDER BY entry_number")
     @RegisterMapper(EntryMapper.class)
-    @FetchSize(262144) // Has to be non-zero to enable cursor mode https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
+    @FetchSize(10000) // Has to be non-zero to enable cursor mode https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
```
michaelabenyohai: Just wondering why you changed this to 10000? Is it arbitrary? Only because this particular query is used to for proofs so I was wondering if it has any effect on the proof calculations?

PR #175, opened 2017-01-05
Make similar registers available to home page template from config file;

karlbaker02:
This PR is very similar to the one merged a few days ago, this time making available from the config file an optional list of similar registers to the `HomePageView`.

### DIFF
--- a/src/main/java/uk/gov/register/configuration/RegisterContentPages.java
+++ b/src/main/java/uk/gov/register/configuration/RegisterContentPages.java
@@ -1,14 +1,17 @@
 package uk.gov.register.configuration;
 
+import java.util.List;
 import java.util.Optional;
 
 public class RegisterContentPages {
     private final Optional<String> registerHistoryUrl;
     private final Optional<String> custodianName;
+    private final Optional<List<String>> similarRegisters;
 
-    public RegisterContentPages(Optional<String> registerHistoryUrl, Optional<String> custodianName) {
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/register/configuration/RegisterContentPages.java
```diff
@@ -1,14 +1,17 @@
...
+import java.util.List;
 import java.util.Optional;
 
 public class RegisterContentPages {
     private final Optional<String> registerHistoryUrl;
     private final Optional<String> custodianName;
+    private final Optional<List<String>> similarRegisters;
 
-    public RegisterContentPages(Optional<String> registerHistoryUrl, Optional<String> custodianName) {
+    public RegisterContentPages(Optional<String> registerHistoryUrl, Optional<String> custodianName, Optional<List<String>> similarRegisters) {
```
michaelabenyohai: Could you rename `RegisterContentPages.java` to something else? It was originally called "RegisterContentPages" because it contained links to the various content pages related to registers - i.e. the register history page, the spec and other registers guidance but it now contains things other than just "register content pages".

karlbaker02: Should we rename this or should we create a separate class for the things which aren't links to content pages?

michaelabenyohai: I don't mind them being in the same class. I guess they are all things only required by the home page so they have that in common.

PR #176, opened 2017-01-06
Calculate links to from register

karlbaker02:
This PR adds functionality to calculate links between all registers, exposed through a new `RegisterLinkService` and makes this data available to `HomePageView`. Links between registers will be recalculated when the main register application starts, in addition to on demand through the `/delete-register-data` endpoint when register and field configurations are re-loaded.


johnollier:
I've checked out this branch and i'm having a look.

karlbaker02:
@johnollier In the same way that we access the `RegistersConfiguration` and `FieldsConfiguration` in the application, I thought it best to follow a similar here. We could pass in `RegisterLinks` to the constructor of `PostgresRegister` if you think this would be a better way of doing things?

robmckinnon:
@karlbaker02  Does the linked from list include curie links from other registers?

karlbaker02:
@robmckinnon No, curies are excluded for both the 'Links To' and 'Links From' sections.

PR #177, opened 2017-01-10
Add Custodian to register home page

danmitchell-:
## Before
screen shot 2017-01-10 at 10 25 59

## After
screen shot 2017-01-10 at 10 24 55

This commit also removes colons from sidebar titles


karlbaker02:
👍

PR #179, opened 2017-01-10
Remove more gov.uk styles

danmitchell-:
- Fix font so that we don't reference `nta` at all
- Remove blue underline from header
- Remove blue line from top of right hand column
- Update header background colour to proper blue
- Minor update to right column 'last updated' HTML

## Before
screen shot 2017-01-10 at 14 24 54

## After
screen shot 2017-01-10 at 14 23 56


philandstuff:
:shipit:

PR #181, opened 2017-01-11
Simplify similar registers

philandstuff:
Changes an `Optional<List<?>>` to a plain `List<?>`, and simplifies a conditional in a template.


karlbaker02:
Observed Phil making these changes, all is good.

PR #183, opened 2017-01-12
Move fragments into fragments directory

danmitchell-:
Moving all fragments from outside the fragments directory and main.html and into separate files inside the fragments directory.

Also fixed some line height issues with buttons and the phase banner because we aren't using the NTA Transport font.


philandstuff:
in what way is this "WIP"? looks finished to me

PR #185, opened 2017-01-12
Fix header for really long register names

danmitchell-:
### Before
screen shot 2017-01-12 at 15 53 58

### After
screen shot 2017-01-12 at 15 52 47


tvararu:
Looks good. 👍

PR #187, opened 2017-01-12
Add dependency-check gradle plugin

philandstuff:
this adds the [OWASP dependency check][] gradle plugin.  You can run
it using:

    ./gradlew dependencyCheck

I have also made some changes based on its reports:

 - I deliberately exclude the `jrubyExec` configuration because it
   doesn't go into the final app artefact; it's only used to compile
   sass
 - there is a dependency-check-suppressions.xml file which describes
   false positives to suppress.  In particular, markdownj-core was
   falsely detected as vulnerable to CVE-2012-4232 (which instead
   affects jCore), so this has been suppressed.
 - there was an alarm (probably false?) for
   jackson-dataformat-cbor:2.6.6 due to a transitive dependency from
   AWS SDK.  I've added a thing to force it to the same version of
   jackson as we use everywhere else.

[OWASP dependency check]: http://jeremylong.github.io/DependencyCheck/index.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,7 @@ compass {
 }
 
 dependencyCheck {
+    failBuildOnCVSS = 0
     skipConfigurations = ['jrubyExec']
     suppressionFile = 'dependency-check-suppressions.xml'
 }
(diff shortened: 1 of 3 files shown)


michaelabenyohai:
@philandstuff please can you rebase? Thanks!

philandstuff:
conflicts now resolved

karlbaker02:
👍

PR #189, opened 2017-01-13
Update field description layout on home page

danmitchell-:
Update un-ordered list to definition list for the fields name

### Before
screen shot 2017-01-13 at 15 52 53

### After
screen shot 2017-01-16 at 09 19 11


tvararu:
Looks good to me, checked it on mobile on your machine. I think it's good to merge on green. 👍

PR #196, opened 2017-01-19
Fields location in config

johnollier:
NOT for merging until new config items have been added for test etc. Just for review.

I've added the ability to read classpath://xxx urls and then set the default config items in the project to be classpath urls. This is to reduce the number of Optional objects being passed around when the config is read as the application will complain and fail to start if the config items are missing. I also tidied up a couple of tests.

### DIFF
--- a/src/main/java/uk/gov/register/configuration/ConfigManager.java
+++ b/src/main/java/uk/gov/register/configuration/ConfigManager.java
@@ -12,9 +12,9 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class ConfigManager {
-    private final Optional<String> registersConfigFileUrl;
+    private final String registersConfigFileUrl;
     private final String registersConfigFilePath;
-    private final Optional<String> fieldsConfigFileUrl;
+    private final String fieldsConfigFileUrl;
     private final String fieldsConfigFilePath;
     private final boolean refresh;
     private final String externalConfigDirectory;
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/register/configuration/ConfigManager.java
```diff
@@ -57,11 +52,11 @@ public FieldsConfiguration getFieldsConfiguration() {
         return fieldsConfiguration.get();
     }
 
-    private RegistersConfiguration createRegistersConfiguration() {
-        return new RegistersConfiguration(registersConfigFileUrl.map(s -> registersConfigFilePath));
+    private RegistersConfiguration createRegistersConfiguration() throws IOException {
```
karlbaker02: Should this throw `IOException` when the corresponding `createFieldsConfiguration` method doesn't?

src/main/java/uk/gov/register/protocols/classpath/ClasspathURLConnection.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ClasspathURLConnection extends URLConnection {
+
+    private final Pattern classpathPattern = Pattern.compile("classpath://(.+)");
+
+    ClasspathURLConnection(URL url) {
```
karlbaker02: Protected constructor?

johnollier: It only gets called from uk.gov...protocols.classpath.Handler

src/main/java/uk/gov/register/configuration/ConfigManager.java
johnollier: You're right.

chr1sdi:
👍

PR #198, opened 2017-01-23
Remove orphan entry logic that also exists in parser

michaelabenyohai:
This removes logic that checks for entries being appended to a register
that do not have an item in the register. This code currently causes the execution
of an RSF file to read from the database alot, which slows loading down. Since this
logic is now replicated in the RSFExecutor (without going to the database too much) I
am removing it from PostgresRegister. This change assumes that loading data via RSF is
the only way that entries can be appended to a register.


chr1sdi:
👍

PR #199, opened 2017-01-24
Simplify itemContent method logic

michaelabenyohai:
The logic in this method was duplicated in ItemCoverter.convertItem()
so it seemed sensible to reuse this.


chr1sdi:
👍

PR #200, opened 2017-01-26
Improve client exception handling

chr1sdi:
https://trello.com/c/GEFbk8hj/597-improve-exception-mapping-stop-mapping-lots-of-exceptions-to-500


philandstuff:
can your commit messages:

 - explain more of the "why" of the changes rather than just the "what"
 - not have such a long summary line that github truncates them (if you need more, write paragraphs below the summary line)
 - have enough context for them to make sense on their own?

Linking to the trello card is nice, but code often lives longer than project management tools and so in two years when I `git log -p -S exceptionView` to find these commits, I won't understand your intention when you introduced this code 😕

src/main/java/uk/gov/register/providers/ClientErrorExceptionMapper.java
```diff
@@ -35,13 +32,18 @@ public Response toResponse(ClientErrorException exception) {
...
         }
 
         if (exception instanceof NotFoundException || exception instanceof NotAcceptableException) {
-            return viewFactory.thymeleafView("404.html");
+            return viewFactory.exceptionNotFoundView();
         }
 
-        throw exception;
+        if (exception instanceof NotAllowedException) {
+            return viewFactory.exceptionView("Method not allowed", "This method is not allowed");
```
philandstuff: when do you decide to have a custom method like `exceptionNotFoundView()`, and when do you call the generic `exceptionView` with parameters like this?

src/test/java/uk/gov/register/functional/ApplicationTest.java
```diff
@@ -82,10 +83,27 @@ public void appSupportsXssProtection() throws Exception {
...
+    @Test
+    public void returns405ForNotAllowedMethod() {
+        Response response = register.getRequest(address, "/load-rsf");
+
+        assertThat(response.getStatus(), equalTo(405));
+    }
+
+    @Test
+    public void returns400BadRequest() {
+        Response response = register.authenticatedTarget(address).path("/load-rsf").request()
```
philandstuff: why not use `register.loadRsf("nope")` here? it's simpler and we don't need to make `authenticatedTarget` public.

src/main/java/uk/gov/register/providers/ClientErrorExceptionMapper.java
chr1sdi: Either if it appears more than once or it is very common like 400 which can lead to it being used multiple times in the future :)

src/test/java/uk/gov/register/functional/ApplicationTest.java
chr1sdi: I knew you were going to pick on that :D heh there is a test already `DataUploadTest` which uses
`register.mintLines` to cause that exception. My intention here was to be more explicit when it comes to what is actually happening and exposing how I can achieve 400 without investigating methods like `register.loadRsf("nope")`. But fair enough it doesn't look less informative when changed to `loadRsf`. Will change that.

PR #201, opened 2017-01-26
move to official openjdk docker image

philandstuff:
This moves us to the official openjdk docker image, rather than
jstepien/openjdk8.  Currently, the official build is at `u111` while
jstepien/openjdk8 is `u92`.

Also, I picked the alpine-based image.  Alpine is a super minimal
linux environment suitable for running a single process, ideal for our
use case.  (The `openjdk:8-jre` image is based on debian.)


philandstuff:
I tested this manually on multi-1.test.openregister, it works a treat

PR #202, opened 2017-01-26
Refresh register metadata after deleting data fix

chr1sdi:
https://trello.com/c/nkoekkqu/596-register-metadata-fails-to-update-in-some-places-upon-delete-register-data

This fixes the update of the fields. If a user modifies register or field
data and then resets a register (eg by calling `/delete-register-data`)
the register metadata is not being refreshed making home page look the same etc.

The bug was in `RegisterContext` class which was storing `RegisterMetadata`
as field passed in the constructor hence after a reload even though
`ConfigManage`r  had never data Context was still using it's field.

### DIFF
--- a/src/test/java/uk/gov/register/core/RegisterContextTest.java
+++ b/src/test/java/uk/gov/register/core/RegisterContextTest.java
@@ -57,7 +57,7 @@ public class RegisterContextTest {
 
     @Test
     public void getRegisterMetadata_returnsUpToDateConfigProvidedByConfigManager() {
-        RegisterMetadata initialMetadata = new RegisterMetadata(
+        RegisterMetadata expectedInitialMetadata = new RegisterMetadata(
                 new RegisterName("test-register-1"),
                 Collections.emptyList(),
                 "copyright-1",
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/register/core/RegisterContextTest.java
```diff
@@ -50,4 +54,40 @@ public void resetRegister_shouldResetRegister_whenEnableRegisterDataDeleteIsEnab
...
+        when(rcMock.getRegisterMetadata(registerName)).thenReturn(initialMetadata);
+        when(configManager.getRegistersConfiguration()).thenReturn(rcMock);
+
+        RegisterContext context = new RegisterContext(registerName, configManager, registerLinkService, dbi, flyway, Optional.empty(), true, false, Optional.empty(), Optional.empty(), emptyList(), new RegisterAuthenticator("", ""));
+
+        RegisterMetadata actualInitialMetadata = context.getRegisterMetadata();
+        assertThat(actualInitialMetadata, equalTo(initialMetadata));
+
+        when(rcMock.getRegisterMetadata(registerName)).thenReturn(expectedUpdatedMetadata);
+        when(configManager.getRegistersConfiguration()).thenReturn(rcMock);
```
michaelabenyohai: Instead of repeating these two lines, I think you could use `when(rcMock.getRegisterMetadata(registerName)).thenReturn(initialMetadata).thenReturn(expectedUpdatedMetadata)`. Removes a tiny bit of code :).

michaelabenyohai:
I've tested this and it now works as expected. Happy to merge once we've had a chat about my comment on the test 👍 .

PR #205, opened 2017-01-30
Add breadcrumbs to all views

danmitchell-:
### After

breadcrumbs


src/main/resources/assets/sass/_breadcrumbs.scss
```diff
@@ -0,0 +1,19 @@
...
+.breadcrumbs {
+  @include breadcrumbs;
+
+  ol {
+    margin: 0;
+    padding: 0;
+  }
+
+  li {
+    background-image: image-url("images/seperator.png");
```
tvararu: Looks like you got a typo in the image name, should be `separator.png`.

src/main/resources/assets/sass/_breadcrumbs.scss
```diff
@@ -0,0 +1,19 @@
...
+
+    @include device-pixel-ratio() {
```
tvararu: I haven't seen this mixin before but I assume it means "retina" when used without arguments.

PR #208, opened 2017-02-02
Make fields description available on pages (records, record, item)

chr1sdi:
Additionally extracted fields html fragment which can be re-used on multiple pages, sample from homepage:
`<div th:replace="fragments/field-list.html::field-list(${fields})"></div>`

story: https://trello.com/c/SwsyrCzo/619-make-field-descriptions-available-to-records-record-and-item-pages


karlbaker02:
Looking good

PR #212, opened 2017-02-15
support compassWatch properly

philandstuff:
This configures gradle properly such that

    ./gradlew compassWatch

will start a task which watches for changes in .scss files and writes
them into the correct path, such that a running app (started with
./run-application.sh) will see the updated css files.

This means we can iterate on scss changes without needed to restart
the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,3 @@ deployable_bundle
 .cache
 
 .sass-cache
-src/main/resources/assets/css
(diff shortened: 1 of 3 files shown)


danmitchell-:
LOVE IT CHANGE 😍

philandstuff:
would appreciate a backend developer review, just to ensure I haven't messed anything up in gradle

karlbaker02:
It looks fine to me, though I'm by no means a gradle expert.

PR #216, opened 2017-02-28
Detect if running on register gov uk and remove branding if not

danmitchell-:


src/main/resources/templates/download.html
```diff
@@ -2,7 +2,7 @@
 
 <html xmlns:th="http://www.thymeleaf.org" xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
 <head th:include="fragments/head.html::head"></head>
-<body>
+<body th:class="${isGovukBranded ? 'gov-uk' : 'openregister-org'}">
```
tvararu: Is it easy to add a helper that DRYes this? Something like:

```html
<body th:class="${getBrandingClassName()}">
```

```java
// IANAJD
public Boolean getIsGovukBranded() {
    return getIsGovukBranded()
        ? 'gov-uk'
        : 'openregister-org';
}
```

danmitchell-: oooo this would be DRYer 👍

chr1sdi: DRY --> increases coupling

philandstuff: the javabean-convention magic means that you would be able to just say `brandingClassName` in the template and it will get rewritten to `getBrandingClassName()`

danmitchell-: Yep, I have done this just trying to add a test for it 🤓

chr1sdi:
CSS class names in java page object :/ a bit leaky isn't it?

src/main/resources/templates/download.html
danmitchell-: Decided against this as the team doesn't want the java to know about CSS classes 👍

src/main/java/uk/gov/register/thymeleaf/ThymeleafView.java
```diff
@@ -94,6 +94,10 @@ public String getScheme() {
         return requestContext.getScheme();
     }
 
+    @SuppressWarnings("unused, used by templates")
+    public Boolean getIsGovukBranded() {
+        return requestContext.getHost().endsWith("register.gov.uk");
```
philandstuff: this will break if the Host header includes the port number (eg `Host: country.register.gov.uk:443`)

philandstuff: discussed IRL with @samcrang - this is probably fine for now, especially since this is a temporary solution.

the [host header definition](https://tools.ietf.org/html/rfc7230#section-5.4) says that the port should be specified if it's not the default port, but doesn't say if it might be specified if it is the default port.

PR #217, opened 2017-03-02
Add high res organisation logos for retina displays

danmitchell-:
You can just about see the difference ;-)

## Before
screen shot 2017-03-02 at 14 41 38

## After
screen shot 2017-03-02 at 14 41 20


samcrang:
LGTM :shipit:

PR #220, opened 2017-03-08
Add support to Entry for multiple item hashes

karlbaker02:
This PR adds support for multiple items per entry in the register model. It is a fairly straightforward change, but I will elaborate a little on the modification of the Entry `equals` method.

In order to be able to compare items in the entry, we don't need to worry about what order these item hashes are in. We also allow the possibility of duplicate item hashes per entry, so we can't just store all of these hashes as a set. To get around this problem and not need to re-invent the wheel, I've pulled in the Apache Commons Collection library (http://commons.apache.org/proper/commons-collections/index.html) and am using the method `CollectionUtils.isEqualCollection` to satisfy the above requirements (see the source code at https://commons.apache.org/proper/commons-collections/apidocs/src-html/org/apache/commons/collections4/CollectionUtils.html). If we didn't pull in this library and instead wanted to use the default `List.equals` method that Java provides, we'd have to sort the hashes stored by each entry before doing comparison, as this is how the default method operates. If/when we get to the scale where each entry has many items, this may not be feasible, and is a reason why I went with the Commons Collection library (this by comparison builds up a map of frequencies of each of the items in each of the lists, eliminating the need to sort, but still resulting in an entire pass of each of the lists on each comparison).

Pulling in this library raised a potential security vulnerability during our build process (https://tools.cisco.com/security/center/content/CiscoSecurityAdvisory/cisco-sa-20151209-java-deserialization) but after studying this with Sam, I don't believe this will affect us as we don't allow users to supply serialized data.

Aside from the above, this PR also changes the `equals` and `hashcode` methods of `Entry` to enforce that the fields `key` and `timestamp` should be taken into account when comparing the equality of entries; we were not doing this before. As a consequence, some of the tests have been refactored to accommodate for this.

### DIFF
--- a/src/main/java/uk/gov/register/core/Entry.java
+++ b/src/main/java/uk/gov/register/core/Entry.java
@@ -10,12 +10,14 @@ import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
 import com.fasterxml.jackson.dataformat.csv.CsvSchema;
 import com.google.common.collect.Lists;
+import org.apache.commons.collections4.CollectionUtils;
 import uk.gov.register.util.HashValue;
 import uk.gov.register.util.ISODateFormatter;
 import uk.gov.register.views.CsvRepresentationView;
 import uk.gov.register.views.representations.CsvRepresentation;
 
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
@@ -30,7 +32,7 @@ public class Entry implements CsvRepresentationView<Entry> {
 
     public Entry(int entryNumber, HashValue hashValue, Instant timestamp, String key) {
         this.entryNumber = entryNumber;
-        this.hashValues = Arrays.asList(hashValue);
+        this.hashValues = new ArrayList<>(Arrays.asList(hashValue));
         this.timestamp = timestamp;
         this.key = key;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/register/core/Entry.java
```diff
@@ -16,20 +16,28 @@
...
     private final int entryNumber;
-    private final HashValue hashValue;
+    private final List<HashValue> hashValues;
     private final Instant timestamp;
     private String key;
 
     public Entry(int entryNumber, HashValue hashValue, Instant timestamp, String key) {
         this.entryNumber = entryNumber;
-        this.hashValue = hashValue;
+        this.hashValues = Arrays.asList(hashValue);
```
johnollier: You won't be able to add to the List returned by Arrays.asList(). We should create a new instance of List here.

karlbaker02: Java. I'll change this.

michaelabenyohai:
Can we add some tests for the `equals()` and `hashcode()` methods to `EntryTests`? 

Like - do we consider entries that have the same items but in a different order return `true` for `equals()`? We should work that out and add tests to validate the decision.

michaelabenyohai:
We also need to make sure it's possible for an entry to point to the same item twice (or more). For example, it's possible for more than one entry in the original register to point to the same item, which means that it's possible for an entry in a derivation to point to the same item-hash twice.

Can you add a test that makes sure it's possible for an entry to contain the same item-hash multiple times?

PR #222, opened 2017-03-10
Parse multi item entries

johnollier:
This enables the loading of RSF where entry commands may refer to >1 item hash. Note that the database work is still to be done, so while the loading does work, Entries in the database will only refer to the first Item. Item hashes are delimited by ";" in new look RSF. There is an example in src/test/resources/fixtures/serialized/local-auth-by-type-10.rsf for the 'local authorities by type' derivation. When loading, config.yaml should provide for the base register to be 'local-authority-eng' and the fields and registers yaml should be taken from Discovery.


src/main/java/uk/gov/register/serialization/RSFExecutor.java
```diff
@@ -70,22 +72,23 @@ private RegisterResult validate(RegisterCommand command, Register register, int
...
+
+        hashes.stream().forEach(hashValue -> {
+            if (hashRefLine.containsKey(hashValue)) {
+                hashRefLine.put(hashValue, 0);
             }
-        }
-        return false;
+        });
+
+        return hashes.stream().allMatch(hashValue -> (!hashRefLine.containsKey(hashValue)
```
michaelabenyohai: It feels like we are streaming hashes twice here, possibly unnecessarily. Can we loop it once and use the existing logic?

```
for (HashValue hashValue : hashes) {
    if (hashRefLine.containsKey(hashValue)) {
        hashRefLine.put(hashValue, 0);
    } else {
        Optional<Item> item = register.getItemBySha256(hashValue);
        if (!item.isPresent()) {
            return true;
        }
    }
}
return false;

```

michaelabenyohai: With derivations, we're also going to get a lot of repeating item-hashes I think. So if we see a item-hash that is not in the RSF, but is in the database, can we store that we've seen it in memory? That way we won't have to go to the database more than necessary. Maybe something like:

```
for (HashValue hashValue : hashes) {
    if (hashRefLine.containsKey(hashValue)) {
        hashRefLine.put(hashValue, 0);
    } else {
        Optional<Item> item = register.getItemBySha256(hashValue);
        if (!item.isPresent()) {
            return true;
        } else {
            hashRefLine.put(hashValue, 0);
        }
    }
}
return false;
```

src/test/resources/fixtures/serialized/local-auth-by-type-10.rsf
```diff
@@ -0,0 +1,13 @@
+add-item	{"local-authority-eng":"BAS","local-authority-type":"UA","name":"Bath and North East Somerset","official-name":"Bath and North East Somerset Council","start-date":"1996-04-01"}
```
michaelabenyohai: I can't find where this file is used?

johnollier: I thought it was useful to have an example of multi item RSF in the project. When the other work is done, I was going to use the example in a functional test - hence the location. Perhaps we should allow comments in RSF!

src/main/java/uk/gov/register/serialization/RSFExecutor.java
johnollier: Re throwing an exception - we could do that if we wanted to merge the parsing code into master, but I don't think we will do that before the rest of the derivations code is done.

Re streaming twice - that method was already not very nice in that it does two things at the same time ...

michaelabenyohai:
Another thought. Could you add some tests to make sure that `append-entry` with and empty list of item-hashes works correctly?

This would look like `append-entry\ttimestamp\t\tkey` - do you think it's ok that an entry with no item-hashes is represented by an empty column in TSV? It might be difficult to see.

michaelabenyohai:
This fails to load - I believe because of the `append-entry` that has an empty list of items:

```
add-item	{"citizen-names":"Afghan","country":"AF","name":"Afghanistan","official-name":"The Islamic Republic of Afghanistan"}
add-item	{"citizen-names":"Soviet citizen","country":"SU","name":"USSR","official-name":"Union of Soviet Socialist Republics"}
add-item	{"citizen-names":"Algerian","country":"DZ","name":"Algeria","official-name":"The People's Democratic Republic of Algeria"}
add-item	{"citizen-names":"Albanian","country":"AL","name":"Albania","official-name":"The Republic of Albania"}
add-item	{"citizen-names":"Briton;British citizen","country":"GB","name":"United Kingdom","official-name":"The United Kingdom of Great Britain and Northern Ireland"}
append-entry	2016-04-05T13:23:05Z	sha-256:10d79c864d2c3f2ea6bc8e73d1ac40f5add153a5180bd1ef3a276c877dffc912	SU
append-entry	2016-04-05T13:23:05Z		SU
append-entry	2016-04-05T13:23:05Z	sha-256:6b18693874513ba13da54d61aafa7cad0c8f5573f3431d6f1c04b07ddb27d6bb	GB
append-entry	2016-04-05T13:23:05Z	sha-256:6bf7f01f268fa6d18e53eb7d5ebadb41c25c4aa2eedecfb7bc863e233ec99e24	AF
append-entry	2016-04-05T13:23:05Z	sha-256:9d04a7e04ac92ab809a7471e7142617bc30a4b2d2f30c1002520a4a1d216f2a5	AL
append-entry	2016-04-05T13:23:05Z	sha-256:3548cdf528e450626113d44cdd4ac3a19ff992313fe814899c3198d4e0baeff9	DZ
```

johnollier:
We can allow a representation of an empty list of hashes be it an empty string or something else, but there will still be a problem when we try to store the entry as the mapper calls getSha256 (or whatever). Once the database changes are done, we could come back to this.

johnollier:
I've added the ability to parse append entry command which refers to no item hashes. This feature and parsing multi item entries is only enabled if System Property "multi-item-entries-enabled" is "true".

PR #223, opened 2017-03-13
Migration to add new entry_item table

karlbaker02:
This PR adds a migration to create a new table `entry_item` and insert into it using data currently stored in `entry`. Specifically, it is a join table linking an entry's `entry_number` to an item's `sha256hex`, and this version does not remove `sha256hex` from the entry table - this will happen in a future version.

I've also refactored some of the existing SQL queries to remove stuff like `select`s within `select`s to instead use joins, which should give us better performance and hopefully make it easier to understand what's going on.

`ItemQueryDAO` has also had to change to reflect the addition of the `entry_item` table, and so at the moment `getIterator()` has a `distinct` in the query to only return the unique set of items, rather than the possibility of returning duplicates if an entry references the same item twice or more. This may need to be removed when `download-rsf` supports multiple items per entry.

### DIFF
--- a/src/main/java/uk/gov/register/db/EntryItemDAO.java
+++ b/src/main/java/uk/gov/register/db/EntryItemDAO.java
@@ -1,12 +1,12 @@
 package uk.gov.register.db;
 
+import org.skife.jdbi.v2.sqlobject.BindBean;
 import org.skife.jdbi.v2.sqlobject.SqlBatch;
 import org.skife.jdbi.v2.sqlobject.customizers.BatchChunkSize;
-import uk.gov.register.core.Entry;
-import uk.gov.register.store.postgres.BindEntry;
+import uk.gov.register.util.EntryItemPair;
 
 public interface EntryItemDAO {
-    @SqlBatch("insert into entry_item(entry_number, sha256hex) values (:entry_number, :sha256hex)")
+    @SqlBatch("insert into entry_item(entry_number, sha256hex) values (:entryNumber, :sha256hex)")
     @BatchChunkSize(1000)
-    void insertInBatch(@BindEntry Iterable<Entry> entries);
+    void insertInBatch(@BindBean Iterable<EntryItemPair> entries);
 }
--- a/src/main/java/uk/gov/register/db/EntryItemDAO.java
+++ b/src/main/java/uk/gov/register/db/EntryItemDAO.java
@@ -1,0 +1,12 @@
+package uk.gov.register.db;
+
+import org.skife.jdbi.v2.sqlobject.SqlBatch;
+import org.skife.jdbi.v2.sqlobject.customizers.BatchChunkSize;
+import uk.gov.register.core.Entry;
+import uk.gov.register.store.postgres.BindEntry;
+
+public interface EntryItemDAO {
+    @SqlBatch("insert into entry_item(entry_number, sha256hex) values (:entry_number, :sha256hex)")
+    @BatchChunkSize(1000)
+    void insertInBatch(@BindEntry Iterable<Entry> entries);
+}
--- a/src/main/java/uk/gov/register/db/EntryQueryDAO.java
+++ b/src/main/java/uk/gov/register/db/EntryQueryDAO.java
@@ -18,7 +18,7 @@ import java.util.Optional;
 public interface EntryQueryDAO {
     @RegisterMapper(EntryMapper.class)
     @SingleValueResult(Entry.class)
-    @SqlQuery("select entry_number, sha256hex, timestamp, key from entry where entry_number=:entryNumber")
+    @SqlQuery("select e.entry_number, ei.sha256hex, e.timestamp, e.key from entry e join entry_item ei on ei.entry_number = e.entry_number where e.entry_number = :entryNumber")
     Optional<Entry> findByEntryNumber(@Bind("entryNumber") int entryNumber);
 
     @RegisterMapper(LongTimestampToInstantMapper.class)
@@ -31,24 +31,24 @@ public interface EntryQueryDAO {
 
     //Note: This is fine for small data registers like country
     @RegisterMapper(EntryMapper.class)
-    @SqlQuery("SELECT entry_number, sha256hex, timestamp, key from entry order by entry_number desc")
+    @SqlQuery("SELECT e.entry_number, ei.sha256hex, e.timestamp, e.key from entry e join entry_item ei on ei.entry_number = e.entry_number order by e.entry_number desc")
     Collection<Entry> getAllEntriesNoPagination();
 
     @RegisterMapper(EntryMapper.class)
-    @SqlQuery("select entry_number, sha256hex, timestamp, key from entry where entry_number >= :start and entry_number \\< :start + :limit order by entry_number asc")
+    @SqlQuery("select e.entry_number, ei.sha256hex, e.timestamp, e.key from entry e join entry_item ei on ei.entry_number = e.entry_number where e.entry_number >= :start and e.entry_number \\< :start + :limit order by e.entry_number asc")
     Collection<Entry> getEntries(@Bind("start") int start, @Bind("limit") int limit);
 
-    @SqlQuery("SELECT entry_number, sha256hex, timestamp, key FROM entry WHERE entry_number >= :entryNumber ORDER BY entry_number")
+    @SqlQuery("SELECT e.entry_number, ei.sha256hex, e.timestamp, e.key FROM entry e join entry_item ei on ei.entry_number = e.entry_number WHERE e.entry_number >= :entryNumber ORDER BY e.entry_number")
     @RegisterMapper(EntryMapper.class)
     @FetchSize(262144) // Has to be non-zero to enable cursor mode https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
     ResultIterator<Entry> entriesIteratorFrom(@Bind("entryNumber") int entryNumber);
 
-    @SqlQuery("SELECT entry_number, sha256hex, timestamp, key FROM entry ORDER BY entry_number")
+    @SqlQuery("SELECT e.entry_number, ei.sha256hex, e.timestamp, e.key FROM entry e join entry_item ei on ei.entry_number = e.entry_number ORDER BY e.entry_number")
     @RegisterMapper(EntryMapper.class)
     @FetchSize(10000)
     Iterator<Entry> getIterator();
 
-    @SqlQuery("SELECT entry_number, sha256hex, timestamp, key FROM entry WHERE entry_number > :totalEntries1 and entry_number <= :totalEntries2 ORDER BY entry_number")
+    @SqlQuery("SELECT e.entry_number, ei.sha256hex, e.timestamp, e.key FROM entry e join entry_item ei on ei.entry_number = e.entry_number WHERE e.entry_number > :totalEntries1 and e.entry_number <= :totalEntries2 ORDER BY e.entry_number")
     @RegisterMapper(EntryMapper.class)
     @FetchSize(10000)
     Iterator<Entry> getIterator(@Bind("totalEntries1") int totalEntries1, @Bind("totalEntries2") int totalEntries2);
--- a/src/main/java/uk/gov/register/db/EntryQueryDAO.java
+++ b/src/main/java/uk/gov/register/db/EntryQueryDAO.java
@@ -18,7 +18,7 @@ import java.util.Optional;
 public interface EntryQueryDAO {
     @RegisterMapper(EntryMapper.class)
     @SingleValueResult(Entry.class)
-    @SqlQuery("select e.entry_number, ei.sha256hex, e.timestamp, e.key from entry e join entry_item ei on ei.entry_number = e.entry_number where e.entry_number = :entryNumber")
+    @SqlQuery("select e.entry_number, array_remove(array_agg(ei.sha256hex), null) as sha256hex, e.timestamp, e.key from entry e left join entry_item ei on ei.entry_number = e.entry_number where e.entry_number = :entryNumber group by e.entry_number")
     Optional<Entry> findByEntryNumber(@Bind("entryNumber") int entryNumber);
 
     @RegisterMapper(LongTimestampToInstantMapper.class)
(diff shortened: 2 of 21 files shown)


michaelabenyohai:
This doesn't quite work at the moment. If the entry_item table contains 2 rows against entry_number 1, the queries in `EntryQueryDAO` return two entries with entry-number 1.

E.g. With the following in the table:

screen shot 2017-03-13 at 13 05 39

I get the following in the /entries endpoint:

screen shot 2017-03-13 at 13 04 29

This is because `EntryMapper` works on a per-database-row basis. Maybe we need some aggregation function (https://www.postgresql.org/docs/9.0/static/functions-aggregate.html) though I'm not sure what the performance of these is like, would be worth checking.

src/main/java/uk/gov/register/db/ItemQueryDAO.java
```diff
@@ -24,12 +24,14 @@
     @RegisterMapper(ItemMapper.class)
     Collection<Item> getAllItemsNoPagination();
 
-    @SqlQuery("select sha256hex, content from item where exists(select 1 from entry where item.sha256hex = entry.sha256hex)")
+    // TODO: This query will need to change to remove `distinct` when download-rsf supports multiple items per entry.
+    @SqlQuery("select distinct ei.sha256hex, i.content from item i join entry_item ei on i.sha256hex = ei.sha256hex")
```
michaelabenyohai: I think it would be simpler to use the existing query (but using `entry_item` instead of `entry`). We only care about returning each item once and since they are guaranteed to only appear in the `item` table once, you don't have to worry about using `distinct`. It's also a smaller change.

```
select sha256hex, content from item where exists(select 1 from entry_item ei where item.sha256hex = ei.sha256hex);
```

This is actually more performant:

```
explain select sha256hex, content from item where exists(select 1 from entry_item where item.sha256hex = entry_item.sha256hex);
                                QUERY PLAN
---------------------------------------------------------------------------
 Hash Semi Join  (cost=35.50..87.25 rows=1000 width=191)
   Hash Cond: ((item.sha256hex)::text = (entry_item.sha256hex)::text)
   ->  Seq Scan on item  (cost=0.00..38.00 rows=1000 width=191)
   ->  Hash  (cost=23.00..23.00 rows=1000 width=65)
         ->  Seq Scan on entry_item  (cost=0.00..23.00 rows=1000 width=65)
(5 rows)
```

than:

```
explain select distinct ei.sha256hex, i.content from item i join entry_item ei on i.sha256hex = ei.sha256hex;
                                     QUERY PLAN
------------------------------------------------------------------------------------
 HashAggregate  (cost=92.25..102.25 rows=1000 width=191)
   Group Key: ei.sha256hex, i.content
   ->  Hash Join  (cost=35.50..87.25 rows=1000 width=191)
         Hash Cond: ((i.sha256hex)::text = (ei.sha256hex)::text)
         ->  Seq Scan on item i  (cost=0.00..38.00 rows=1000 width=191)
         ->  Hash  (cost=23.00..23.00 rows=1000 width=65)
               ->  Seq Scan on entry_item ei  (cost=0.00..23.00 rows=1000 width=65)
(7 rows)
```

src/main/java/uk/gov/register/db/ItemQueryDAO.java
```diff
@@ -24,12 +24,14 @@
...
     @RegisterMapper(ItemMapper.class)
     @FetchSize(10000)
     Iterator<Item> getIterator();
 
-    @SqlQuery("select sha256hex, content from item where exists(select 1 from entry where item.sha256hex = entry.sha256hex and entry_number > :startEntryNo and entry_number <= :endEntryNo)")
+    // TODO: This query will need to change to remove `distinct` when download-rsf supports multiple items per entry.
+    @SqlQuery("select distinct ei.sha256hex, i.content from item i join entry_item ei on i.sha256hex = ei.sha256hex and ei.entry_number > :startEntryNo and ei.entry_number <= :endEntryNo")
```
michaelabenyohai: Same as above

src/main/java/uk/gov/register/db/EntryItemDAO.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.skife.jdbi.v2.sqlobject.SqlBatch;
+import org.skife.jdbi.v2.sqlobject.customizers.BatchChunkSize;
+import uk.gov.register.core.Entry;
+import uk.gov.register.store.postgres.BindEntry;
+
+public interface EntryItemDAO {
+    @SqlBatch("insert into entry_item(entry_number, sha256hex) values (:entry_number, :sha256hex)")
+    @BatchChunkSize(1000)
+    void insertInBatch(@BindEntry Iterable<Entry> entries);
```
michaelabenyohai: This will not insert multiple rows if the entry has more than one item-hash. `BindEntry` still calls `getSha256hex` instead of the new `getItemHashes`. Remember also that an entry could have an empty list of hashes, in which case you would not insert anything in this table.

src/main/java/uk/gov/register/db/RecordQueryDAO.java
```diff
@@ -27,20 +27,20 @@
     @SqlQuery("SELECT count FROM total_records")
     public abstract int getTotalRecords();
 
-    @SqlQuery("select entry_number, timestamp, e.sha256hex as sha256hex, key, content from entry e, item i where e.sha256hex=i.sha256hex and e.entry_number = (select entry_number from current_keys where current_keys.key=:key)")
+    @SqlQuery("select e.entry_number, e.timestamp, ei.sha256hex as sha256hex, e.key, i.content from entry e join entry_item ei on ei.entry_number = e.entry_number join item i on i.sha256hex = ei.sha256hex and e.entry_number = (select entry_number from current_keys where current_keys.key=:key)")
```
michaelabenyohai: These queries also only return a single item for each entry (record). They need to be changed to return all the items. This also means that you may need to change `Record` to contain a list of items, and leave the old single-item getters/constructors there for now (similar to what we did for `Entry`).

src/main/java/uk/gov/register/db/EntryQueryDAO.java
```diff
@@ -18,7 +18,7 @@
 public interface EntryQueryDAO {
     @RegisterMapper(EntryMapper.class)
     @SingleValueResult(Entry.class)
-    @SqlQuery("select entry_number, sha256hex, timestamp, key from entry where entry_number=:entryNumber")
+    @SqlQuery("select e.entry_number, array_agg(ei.sha256hex) as sha256hex, e.timestamp, e.key from entry e join entry_item ei on ei.entry_number = e.entry_number where e.entry_number = :entryNumber group by e.entry_number, e.timestamp, e.key")
```
karlbaker02: This is missing a `left join` on `entry_item` - I'll push the fix in a bit.

src/main/java/uk/gov/register/store/postgres/BindEntry.java
```diff
@@ -16,6 +16,7 @@ public Binder build(Annotation annotation) {
             return (Binder<BindEntry, Entry>) (q, bind, arg) -> {
                 q.bind("entry_number", arg.getEntryNumber());
                 q.bind("sha256hex", arg.getSha256hex().getValue());
+                q.bind("sha256hexes", arg.getItemHashes().stream().map(h -> h.getValue()));
```
karlbaker02: This will also be removed in the next push.

src/main/java/uk/gov/register/db/EntryQueryDAO.java
michaelabenyohai: I also don't think you need `e.timestamp` and `e.key` in the `group by`, given that it is already grouping by `e.entry_number` and this is the primary key of the `entry` table.

karlbaker02: Good spot, I thought I'd removed all of these. Thanks!

src/main/java/uk/gov/register/db/RecordQueryDAO.java
```diff
@@ -21,26 +21,25 @@
...
+    @SqlQuery("select e.entry_number, array_agg(ei.sha256hex) as sha256hex, e.timestamp, e.key, array_agg(i.content) as content from entry e join entry_item ei on ei.entry_number = e.entry_number and e.entry_number = (select entry_number from current_keys where current_keys.key=:key) join item i on i.sha256hex = ei.sha256hex group by e.entry_number")
     @SingleValueResult(Record.class)
     @RegisterMapper(RecordMapper.class)
     public abstract Optional<Record> findByPrimaryKey(@Bind("key") String key);
 
-    @SqlQuery("select entry.entry_number, timestamp, entry.sha256hex as sha256hex, entry.key, content from item, entry, current_keys where current_keys.entry_number = entry.entry_number and item.sha256hex=entry.sha256hex order by entry.entry_number desc limit :limit offset :offset")
+    @SqlQuery("select e.entry_number, array_agg(ei.sha256hex) as sha256hex, e.timestamp, e.key, array_agg(i.content) as content from entry e join entry_item ei on ei.entry_number = e.entry_number join item i on i.sha256hex = ei.sha256hex join current_keys ck on ck.entry_number = e.entry_number group by e.entry_number order by e.entry_number desc")
```
michaelabenyohai: I think you need to keep `limit :limit offset :offset` at the end of this query.

karlbaker02: This must have dropped off when changing the queries back, thanks. Pushing shortly.

src/main/java/uk/gov/register/db/RecordQueryDAO.java
```diff
@@ -21,26 +21,25 @@
...
     @RegisterMapper(RecordMapper.class)
     public abstract List<Record> getRecords(@Bind("limit") long limit, @Bind("offset") long offset);
 
-    @SqlQuery("select entry_number, timestamp, sha256hex, key from entry where sha256hex in (select sha256hex from item where (content @> :contentPGobject)) order by entry_number asc")
+    @SqlQuery("select e.entry_number, array_agg(ei.sha256hex) as sha256hex, e.timestamp, e.key from entry e join entry_item ei on ei.entry_number = e.entry_number join item i on i.sha256hex = ei.sha256hex and i.content @> :contentPGobject group by e.entry_number order by e.entry_number asc")
```
michaelabenyohai: This query can be made much simpler now that we have `key` in the `entry` table - we didn't used to have this, which is why we used the jsonb query. Instead we could do:

```
select e.entry_number, array_remove(array_agg(ei.sha256hex), null) as sha256hex, e.timestamp, e.key from entry e left join entry_item ei on ei.entry_number = e.entry_number where e.key = :key group by e.entry_number;
```

src/main/java/uk/gov/register/db/TransactionalEntryLog.java
```diff
@@ -48,7 +50,12 @@ public void checkpoint() {
         if (stagedEntries.isEmpty()) {
             return;
         }
+
+        List<Entry> mappedEntries = new ArrayList<>();
+        stagedEntries.forEach(se -> se.getItemHashes().forEach(h -> mappedEntries.add(new Entry(se.getEntryNumber(), h, se.getTimestamp(), se.getKey()))));
+
         entryDAO.insertInBatch(stagedEntries);
+        entryItemDAO.insertInBatch(mappedEntries);
```
michaelabenyohai: What happens if `mappedEntries` is empty? Does it still attempt to insert?

src/main/java/uk/gov/register/db/RecordQueryDAO.java
```diff
@@ -16,43 +16,36 @@
...
+    @SqlQuery("select e.entry_number, array_agg(ei.sha256hex) as sha256hex, e.timestamp, e.key, array_agg(i.content) as content from entry e join entry_item ei on ei.entry_number = e.entry_number join item i on i.sha256hex = ei.sha256hex join current_keys ck on ck.entry_number = e.entry_number group by e.entry_number order by e.entry_number desc limit :limit offset :offset")
     @RegisterMapper(RecordMapper.class)
     public abstract List<Record> getRecords(@Bind("limit") long limit, @Bind("offset") long offset);
 
-    @SqlQuery("select entry_number, timestamp, sha256hex, key from entry where sha256hex in (select sha256hex from item where (content @> :contentPGobject)) order by entry_number asc")
+    @SqlQuery("select e.entry_number, array_remove(array_agg(ei.sha256hex), null) as sha256hex, e.timestamp, e.key from entry e left join entry_item ei on ei.entry_number = e.entry_number where e.key = :key group by e.entry_number")
```
michaelabenyohai: I think you just need to add the `order by` clause back.

src/main/java/uk/gov/register/db/TransactionalEntryLog.java
```diff
@@ -48,7 +51,12 @@ public void checkpoint() {
         if (stagedEntries.isEmpty()) {
             return;
         }
+
+        List<EntryItemPair> mappedEntries = new ArrayList<>();
```
michaelabenyohai: Can you rename `mappedEntries`?

src/main/java/uk/gov/register/db/mappers/ItemMapper.java
```diff
@@ -23,7 +21,7 @@ public ItemMapper() {
     @Override
     public Item map(int index, ResultSet r, StatementContext ctx) throws SQLException {
         try {
-            return new Item(new HashValue(HashingAlgorithm.SHA256, r.getString("sha256hex")), objectMapper.readValue(r.getString("content"), JsonNode.class));
+            return new Item(objectMapper.readValue(r.getString("content"), JsonNode.class));
```
michaelabenyohai: I don't think this needs to be changed. I think it's better to read the hash from the database than to recompute it.

src/main/java/uk/gov/register/db/mappers/RecordMapper.java
```diff
@@ -1,26 +1,42 @@
...
         this.entryMapper = new EntryMapper();
-        this.itemMapper = new ItemMapper();
+        objectMapper = Jackson.newObjectMapper();
     }
 
     @Override
     public Record map(int index, ResultSet r, StatementContext ctx) throws SQLException {
+        List<Item> items = Arrays.asList((String[]) r.getArray("content").getArray()).stream().map(c -> {
+            try {
+                return new Item(objectMapper.readValue(c, JsonNode.class));
```
michaelabenyohai: Judging by the documentation (https://www.postgresql.org/docs/9.5/static/functions-aggregate.html) the `sha256hex` array and the `content` array should remain in the same order they were read from the joined tables. I think it's therefore safe to loop over both arrays and use the `Item(hash, content)` contructor, rather than recomputing the hash.

karlbaker02: We could put an order by on each of the aggregates if we wanted to be especially safe (https://www.postgresql.org/docs/9.5/static/sql-expressions.html#SYNTAX-AGGREGATES) but I think you're right, that the ordering is based on the order of the input rows.

src/test/java/uk/gov/register/functional/db/TestEntryDAO.java
```diff
@@ -10,33 +10,39 @@
...
+    @SqlQuery("select e.entry_number, array_remove(array_agg(ei.sha256hex), null) as sha256hex, e.timestamp, e.key from entry e left join entry_item ei on ei.entry_number = e.entry_number where e.entry_number >= :entryNumber group by e.entry_number order by e.entry_number")
     @RegisterMapper(EntryMapper.class)
     @FetchSize(262144) // Has to be non-zero to enable cursor mode https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
     ResultIterator<Entry> entriesIteratorFrom(@Bind("entryNumber") int entryNumber);
 
     class EntryMapper implements ResultSetMapper<Entry> {
         @Override
         public Entry map(int index, ResultSet r, StatementContext ctx) throws SQLException {
-            return new Entry(r.getInt("entry_number"), new HashValue(HashingAlgorithm.SHA256, r.getString("sha256hex")), Instant.ofEpochSecond(r.getLong("timestamp")), r.getString("key"));
+            //return new Entry(r.getInt("entry_number"), new HashValue(HashingAlgorithm.SHA256, r.getString("sha256hex")), Instant.ofEpochSecond(r.getLong("timestamp")), r.getString("key"));
```
michaelabenyohai: Can you delete this rather than commenting it out?

PR #224, opened 2017-03-13
Simplify running app locally

samcrang:
Assuming we have people that just want to quickly assess the suitability
of `openregister-java` but are put off because it requires compiling the
app, and installing and configuring Postgres. This simplifies the process
into just one command: `./run-application.sh`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,25 +3,37 @@ A java implementation of a register
 
 [![Build Status](https://travis-ci.org/openregister/openregister-java.svg?branch=master)](https://travis-ci.org/openregister/openregister-java)
 
-# Requirements
+## Running locally
+
+You can spin up a local copy of `openregister-java` using Docker with the following:
+
+    ./run-application.sh
+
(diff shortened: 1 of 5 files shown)


johnollier:
This is great, especially for someone wanting to get started or when we have a separate front end app. I think a back end developer would still want to run the code natively including running the tests. They probably want to be able to poke about in the database using the standard tools as well. It would be good if you could spin up a container running Postgres with the right config to provide admin and app level users and able to accept remote connections. The default config.yaml could point to that.

samcrang:
I agree. At the moment I see this as only making sense for the case where someone wants to get something running locally as soon as possible (in order to play with a register, assess the suitability of `openregister-java`, or whatever).

I think for people developing stuff nothing changes. However, I think we should change to running Postgres inside a container for development separately because I think that might be slightly more objectionable?

samcrang:
Ahh, I understand. I'll split the config files.

samcrang:
@johnollier how's this?

johnollier:
👍

PR #229, opened 2017-03-17
Render multi item entries (DO NOT MERGE)

johnollier:
This changes the way records and entries are rendered in various formats taking account of Entries now referring to multiple Items. These are breaking changes and this PR is just for review. It should not be merged.

Also the build will fail until related changes to the conformance tests have been done.


robmckinnon:
Where can I find documentation of the breaking changes to the API?  It looks like `item-hash` field is to be changed to `item-hashes`.

Is the TSV output for a record going to change? e.g. Currently for https://local-authority-eng.register.gov.uk/record/KIN.tsv the record TSV looks like:

```
entry-number	entry-timestamp	item-hash	local-authority-eng	local-authority-type	name	official-name	start-date	end-date
357	2017-01-26T12:34:10Z	sha-256:3f4da33a33c24de11cca3539f14ee663359608be0ba218d4fc05792c1d19c00f	KIN	NMD	King's Lynn and West Norfolk	Borough Council of King's Lynn and West Norfolk		
```

src/main/java/uk/gov/register/views/RecordView.java
```diff
@@ -35,27 +40,37 @@ public ObjectNode getRecordJson() {
...
         jsonNodes.remove("key");
-        jsonNodes.setAll(objectMapper.convertValue(itemView, ObjectNode.class));
+        ArrayNode items = jsonNodes.putArray("items");
+        itemViews.forEach( iv -> items.add(objectMapper.convertValue(iv, ObjectNode.class)));
         return jsonNodes;
     }
 
-    public Map<String, FieldValue> getContent() {
-        return itemView.getContent();
+    ArrayNode getFlatRecordJson() {
```
karlbaker02: We're introducing some inconsistency here. If you load an entry with two items, here is what you get for the TSV format when you download the record and entry, respectively:

`/record/402160.tsv`:
```entry-number	entry-timestamp	item-hashes	school	name	start-date	end-date	gender	religious-character	minimum-age	maximum-age	headteacher	website	address
3	2016-11-24T17:20:33Z	sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1	402161	First Steps	2007-11-06								10034660829
3	2016-11-24T17:20:33Z	sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04	402160	Options Education Programme	2007-11-06	2010-08-31							100101037877```

`/entry/3.tsv`:
```entry-number	entry-timestamp	item-hashes	key
3	2016-11-24T17:20:33Z	sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1;sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04	402160
```

As you notice, we have the same column `item-hashes` in both outputs, yet in `record` we display each item on the entry as a separate row on the same entry number (effectively rendering  the new `item-hashes` column no more than it's predecessor `item-hash`) and in `entry` we put both hashes into `item-hashes` separated by a `;`.

We should standardise this and choose one approach.

src/main/java/uk/gov/register/core/Entry.java
```diff
@@ -48,11 +48,12 @@ public Instant getTimestamp() {
 
     @SuppressWarnings("unused, used from DAO")
     @JsonProperty("item-hash")
+    @JsonIgnore
```
karlbaker02: Is it only the RDA who are consuming this property? Do we want to make a breaking change or should this be supported alongside the new `item-hashes` for some transitional period?

src/main/resources/templates/records.html
```diff
@@ -3,81 +3,86 @@
...
-              </th:block>
-            </tr>
-          </tbody>
-        </table>
-      </div>
+                </tr>
+                </thead>
+                <tbody>
+                <th:block th:each="record : ${content.records}">
+                    <tr th:each="item : ${record.itemViews}">
```
karlbaker02: There is a conversation we need to have around how we want to display records that have entries that have multiple items. Currently it's going to create a new row in the table for each of the items referenced in the (most recent) entry, which may or may not be confusing for the end user. Perhaps we might want to think about grouping these together under the same key to make it clearer? All of these are just thoughts.

src/main/java/uk/gov/register/views/representations/turtle/EntryTurtleWriter.java
```diff
@@ -36,8 +36,9 @@ protected Model rdfModelFor(Entry entry, boolean includeKey) {
         String entryNumber = Integer.toString(entry.getEntryNumber());
         Resource resource = model.createResource(entryUri(entryNumber).toString())
                 .addProperty(entryNumberProperty, entryNumber)
-                .addProperty(entryTimestampProperty, entry.getTimestampAsISOFormat())
-                .addProperty(itemProperty, model.createResource(itemUri(entry.getSha256hex().encode()).toString()));
+                .addProperty(entryTimestampProperty, entry.getTimestampAsISOFormat());
+
+        entry.getItemHashes().forEach(hash -> resource.addProperty(itemProperty, model.createResource(itemUri(hash.encode()).toString())));
```
karlbaker02: There's a bug here where if an item is included in an entry more than once, it is only written once to the ttl output.

For example:

RSF:
```assert-root-hash	sha-256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
add-item	{"address":"100101030506","end-date":"","name":"Glanaman Home Tution Centre","school":"402175","start-date":"2007-11-07"}
add-item	{"address":"10034660829","end-date":"","name":"First Steps","school":"402161","start-date":"2007-11-06"}
add-item	{"address":"100101037877","end-date":"2010-08-31","name":"Options Education Programme","school":"402160","start-date":"2007-11-06"}
append-entry	2016-11-24T17:20:33Z	sha-256:4e5ece9111ce1024467b4094ac3f477f83ca7ed100e332e06330d607245b0537	402175
append-entry	2016-11-24T17:20:33Z	sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1	402161
append-entry	2016-11-24T17:20:33Z	sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04	402160
append-entry	2016-11-24T17:20:33Z	sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04;sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1;sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04	402160
append-entry	2016-11-24T17:20:33Z	sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04;sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1	402160```

Output from `/entry/4.ttl`:
```@prefix register-metadata:  <https://openregister.github.io/specification/#> .

<http://school.openregister.dev:8080/entry/4>
      register-metadata:entry-number-field
              "4" ;
      register-metadata:entry-timestamp-field
              "2016-11-24T17:20:33Z" ;
      register-metadata:item-resource
              <http://school.openregister.dev:8080/item/sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1> , <http://school.openregister.dev:8080/item/sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04> ;
      register-metadata:key-field
              "402160" .```

Expected:
```@prefix register-metadata:  <https://openregister.github.io/specification/#> .

<http://school.openregister.dev:8080/entry/4>
      register-metadata:entry-number-field
              "4" ;
      register-metadata:entry-timestamp-field
              "2016-11-24T17:20:33Z" ;
      register-metadata:item-resource
              <http://school.openregister.dev:8080/item/sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1> , <http://school.openregister.dev:8080/item/sha-256:cbc12a2ff953c5751b34706990ff32af935942b4901254079457e6e5b5c35c04> , <http://school.openregister.dev:8080/item/sha-256:088bb2124883bf2b82812f6d290cf860d9699fb17368e66ee442c3875162bfa1> ;
      register-metadata:key-field
              "402160" .```

File: src/main/java/uk/gov/register/views/representations/turtle/RecordTurtleWriter.java [hunk 5]
```diff
@@ -31,26 +31,29 @@ public RecordTurtleWriter(javax.inject.Provider<RegisterName> registerNameProvid
...
-        Model itemModel = new ItemTurtleWriter(registerNameProvider, registerResolver).rdfModelFor(itemView);
 
         Resource recordResource = recordModel.createResource(recordUri(view.getPrimaryKey()).toString());
         addPropertiesToResource(recordResource, entryModel.getResource(entryUri(Integer.toString(entry.getEntryNumber())).toString()));
-        addPropertiesToResource(recordResource, itemModel.getResource(itemUri(itemView.getItemHash().encode()).toString()));
 
         Map<String, String> prefixes = entryModel.getNsPrefixMap();
-        prefixes.putAll(itemModel.getNsPrefixMap());
+
+        view.getItemViews().forEach(iv -> {
```
karlbaker02: @michaelabenyohai If the entry for a record has multiple items, and some of those items are duplicates, do we want to show all of the items for the record or just the unique items in the ttl? I suspect the former, as what you would care about when viewing the records is seeing which items belong to it. This is in contrast to when you're viewing an entry, at which time you want t get all of the items, even duplicates.

src/main/resources/templates/records.html
johnollier: True - I just wanted to do something that works at this stage. We might need some help from front end to make it look nice.

src/main/java/uk/gov/register/core/Entry.java
johnollier: This method should be removed after these changes and the database changes are integrated.

src/main/java/uk/gov/register/views/RecordView.java
johnollier: In the entry view we just have the hashes which can be represented in CSV as a delimited string. But in the record view we want the combination of all the Entry and Item properties.

src/main/java/uk/gov/register/views/representations/turtle/RecordTurtleWriter.java
johnollier: Yes - I think the TTL representations are wrong. I'm going to fix it and add some tests for the Turtle writers. (After I've learned Turtle!)

src/main/java/uk/gov/register/core/Entry.java
karlbaker02: That's not exactly what I'm asking, I should have been more explicit. By marking this property with `@JsonIgnore` we're effectively removing it from being included in the download formats, which is the breaking change. Do we want to do this or have a period of time where an entry consists of both `item-hash` and `item-hashes`?

src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -74,7 +74,7 @@ public int getTotalEntries() {
 
     @Override
     public Optional<Record> getRecord(String key) {
-        return recordIndex.getRecord(key);
+        return recordIndex.getRecord(key).filter(record -> !record.getItems().isEmpty());
```
michaelabenyohai: You shouldn't need this change as the filter is performed by the SQL query.

michaelabenyohai:
Some of the resources don't match the new API design All the record resources need to have `item-hash` removed and `key` added as per the new API design https://github.com/openregister/specification/blob/pu/API-changes.txt. The following changes are still required:

- add `key` to /record/key and /records
- remove `item-hash` from /record/key and /records
- /entry/1 needs to be an array to be consistent with /entries
- /record/key needs to be turned into a map object to be consistent with /records

```
{
  "AE": {
    "index-entry-number": "190",
    "entry-number": "190",
    "entry-timestamp": "2016-04-05T13:23:05Z",
    "key": "AE",
    "item":
    [
      {
        "country": "AE",
        "official-name": "The United Arab Emirates",
        "name": "United Arab Emirates",
        "citizen-names": "Citizen of the United Arab Emirates"
      },
      {
        "country": "AE",
        "official-name": "The United Arab Emirates",
        "name": "United Arab Emirates",
        "citizen-names": "Citizen of the United Arab Emirates"
      }
    ]
  }
}
```

src/main/java/uk/gov/register/core/Entry.java
```diff
@@ -19,8 +19,8 @@
 import java.time.Instant;
 import java.util.*;
 
-@JsonIgnoreProperties(ignoreUnknown=true)
-@JsonPropertyOrder({"entry-number", "entry-timestamp", "item-hash", "key"})
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonPropertyOrder({"index-entry-number", "entry-number", "entry-timestamp", "item-hash", "key"})
```
michaelabenyohai: I think we should move `key` to be before `item-hash` so that it doesn't get missed when there are lots of item-hashes.

```
{"index-entry-number", "entry-number", "entry-timestamp", "key", "item-hash"}
```

michaelabenyohai:
The TTL representations are missing the `index-entry-number` property.

src/main/resources/templates/entries.html
```diff
@@ -3,74 +3,76 @@
...
-          </tr>
-          </tbody>
-        </table>
-      </div>
+        <div class="table-wrapper">
+            <table th:if="${#lists.isEmpty(content.entries) == false}">
+                <thead>
+                <tr>
+                    <th scope="col">Entry</th>
+                    <th scope="col">item-hashes</th>
```
michaelabenyohai: This column name needs to remain `item-hash` (and also same for entry-table.html)

src/main/resources/templates/record.html
```diff
@@ -3,56 +3,91 @@
...
+                </details>
             </div>
-          </details>
+            <div class="column-third" th:include="fragments/attribution.html::attribution"></div>
         </div>
-        <div class="column-third" th:include="fragments/attribution.html::attribution"></div>
-      </div>
 
-      <div th:include="fragments/item-table.html :: item-table (content = ${content.content})" class="table-wrapper"></div>
+        <table th:if="${!content.content.isEmpty()}"
```
michaelabenyohai: The spacing between the table and the surrounding crest, field-list etc. looks a bit dodgy. I think we need a `<div class="table-wrapper">` around it like we do in `records.html`.

Also, it looks like your indentation size must be set to 4 for html - previously it has been size 2. I think 2 is the standard for html so please could you make sure you use 2 so that it is easier to see what has actually changed - currently it looks like you have changed every single line and some html files have different size indentation to others.

It looks like `record.html` and `records.html` have duplicate code for the table. Would it be possible to move these out into a commonly used fragment like has been done with `item-table.html`.

PR #233, opened 2017-03-22
Add and test index functions for local authority, current countries.

karlbaker02:
This PR adds some of the functionality that Michaela created in her spike and adds tests for `CurrentCountriesIndexFunction` and `LocalAuthoritiesByTypeIndexFunction`.

### DIFF
--- a/src/main/java/uk/gov/register/indexer/function/LocalAuthorityByTypeIndexFunction.java
+++ b/src/main/java/uk/gov/register/indexer/function/LocalAuthorityByTypeIndexFunction.java
@@ -1,0 +1,25 @@
+package uk.gov.register.indexer.function;
+
+import uk.gov.register.core.Register;
+import uk.gov.register.indexer.IndexValueItemPair;
+import uk.gov.register.util.HashValue;
+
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/register/indexer/function/LocalAuthorityByTypeIndexFunction.java
```diff
@@ -0,0 +1,25 @@
...
+public class LocalAuthorityByTypeIndexFunction extends BaseIndexFunction {
+    private final Register register;
+
+    public LocalAuthorityByTypeIndexFunction(Register register) {
+        this.register = register;
+    }
+
+    @Override
+    protected void execute(String key, HashValue itemHash, Set<IndexValueItemPair> result) {
+        register.getItemBySha256(itemHash).map(i -> result.add(new IndexValueItemPair(i.getValue("local-authority-type"), i.getSha256hex())));
```
michaelabenyohai: We should probably change this logic to check that "local-authority-type" is populated. If it's not populated then we should return an empty set?

PR #234, opened 2017-03-23
Add index table

endamccormack:
Add in index table for derivations.


src/main/resources/sql/V5__Create_index_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS index (
+  name varchar NOT NULL,
+  value varchar NOT NULL,
```
michaelabenyohai: Maybe we should call this `key` - since this is also what we use in the Entry table?

endamccormack: Agreed. Updated.

PR #238, opened 2017-03-31
Derivation download and resource

endamccormack:
Adding new resource to provide record(s) endpoints for derivations. Added to DownloadData resource to add endpoints to download RSF for derivations.  Extend IndexQueryDAO to read records/entries from the index table storing derivations. Wiring up of classes.

### DIFF
--- a/src/main/java/uk/gov/register/serialization/RSFCreator.java
+++ b/src/main/java/uk/gov/register/serialization/RSFCreator.java
@@ -1,6 +1,5 @@
 package uk.gov.register.serialization;
 
-import com.google.common.collect.Iterators;
 import uk.gov.register.core.HashingAlgorithm;
 import uk.gov.register.core.Register;
 import uk.gov.register.util.HashValue;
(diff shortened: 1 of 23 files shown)


src/main/java/uk/gov/register/serialization/RSFCreator.java
```diff
@@ -50,6 +51,30 @@ public RegisterSerialisationFormat create(Register register, int totalEntries1,
...
+        Iterator<RegisterCommand> commands = Iterators.transform(iterators, obj -> (RegisterCommand) getMapper(obj.getClass()).apply(obj));
+        return new RegisterSerialisationFormat(commands);
+    }
+
+
+    public RegisterSerialisationFormat create(Register register, String indexName, int totalEntries1, int totalEntries2) {
+        Iterator<?> iterators;
+
+        if (totalEntries1 == totalEntries2) {
+            iterators = Iterators.singletonIterator(register.getRegisterProof(totalEntries1));
```
michaelabenyohai: I think we need to just return nothing here rather than returning the root hash for the register?

endamccormack: Thats a great catch Michaela. What do you think of `iterators = Collections.<RegisterCommand>emptyList().iterator();` instead?

karlbaker02:
Sat down with John to go through this. Happy for it to be merged once the build passes.

PR #244, opened 2017-04-05
Read config yaml in new format

johnollier:


karlbaker02:
This all looks sensible. Are you going to be merging your conformance tests first? You'll need to update `requirements.txt` here in any case to reference the new commit you make.

PR #247, opened 2017-04-07
Change entry resource to return an array

michaelabenyohai:
This makes the entry and entries resources more consistent and allows us to
possibly return multiple entries from this resource in future should it make sense.

We should refactor the view code at some point soon as this could probably be made simpler.

This requires https://github.com/openregister/conformance-test/pull/24 to be merged first and to update `requirements.txt`.


karlbaker02:
Tested with the updated conformance tests, looks good. Happy for this to be merged once requirements.txt` is updated.

PR #248, opened 2017-04-07
Generate and publish artifact with the same name

samcrang:
This requires [https://github.com/openregister/deployment/pull/329](this) to be deployed first.

CodePipeline seems to require an S3 artifact with a known, constant name. It relies on S3 versioning to reference the different builds of the artifact.

As we're using CodePipeline to trigger CodeDeploy we no longer need to do this from Travis.


samcrang:
After this gets merged we can revoke all of Travis's CodeDeploy permissions.

PR #249, opened 2017-04-07
Use the full key ID for Postgres's Debian repository

samcrang:
This doesn't really add very much but I suppose it prevents against
some theoretical collisions with 16 hex-digit key IDs.


samcrang:
We don't need to do this at all if we merge #250.

PR #253, opened 2017-04-12
Database per deployment group

karlbaker02:
This PR updates the register application to use one database per register group, where each separate register uses its own schema.


src/main/java/uk/gov/register/db/SchemaRewriter.java
```diff
@@ -0,0 +1,13 @@
...
+import org.skife.jdbi.v2.tweak.StatementLocator;
+
+public class SchemaRewriter implements StatementLocator {
+    public static ThreadLocal<String> schema = new ThreadLocal<>();
+
+    @Override
+    public String locate(String name, StatementContext ctx) throws Exception {
+        return ctx.getRawSql().replace(":schema", schema.get());
+    }
+}
```
duchuyctlk: Please add an empty line at the end of file.

PR #257, opened 2017-04-24
Make container build use idiomatic gradle

samcrang:
The `stage` task seems to be related to deploying to Heroku. It adds
nothing to the `assemble` task. The `assemble` task seems to be a more
commonly used task name for assembling a `.jar`.


samcrang:
This depends on #255.

karlbaker02:
There is a good breakdown of the various tasks here: https://docs.gradle.org/current/userguide/java_plugin.html.

Agreed that the `stage` task is a Heroku thing.

PR #259, opened 2017-04-24
Correct link on key column in record(s) views

johnollier:
This is to fix the link on the key column where the record is from an Index. It should now point to  /index/[name]/record/[key] rather than /record/[key]


karlbaker02:
This all looks good.

PR #260, opened 2017-04-24
Read total number of index records for records view.

johnollier:
This adds a query to IndexQueryDao to count the records in a given index. The result shown for e.g. the local-authorities-by-type index is correct but still looks very strange given the way the index records are rendered.


karlbaker02:
You're right in that the page does look strange, but the functional aspect of calculating the correct number of records is what matters for this PR. Hopefully with Blanca's work on the index page we'll solve the display issue and remove any confusion.

src/main/java/uk/gov/register/db/IndexQueryDAO.java
```diff
@@ -237,5 +240,14 @@
             "  index_entry.\"key\"  " +
             "order by  " +
             "  ien  ";
+    
+    String recordCountQuery = "select " +
```
karlbaker02: We could perhaps make this inline rather than create another block for it down here, given that this isn't a particularly complex query like some of the others in this DAO.

johnollier: Even though the query is a lot shorter than the others, I still think its easier to read with line breaks.

PR #262, opened 2017-04-26
Suppress jackson security vulnerability

karlbaker02:
This PR fixes the currently breaking build by suppressing vulnerability `CVE-2016-7051` from the dependency check task. `CVE-2016-7051` regards the use of Jackson's `XmlMapper` class, which is not being used either in ORJ or in DropWizard (the latter also has a dependency on Jackson).

The dependency check report generated by this vulnerability is attached to this PR.
[Dependency-Check-Report.pdf](https://github.com/openregister/openregister-java/files/958486/Dependency-Check-Report.pdf)

Note: the various versions of jackson that are being used will hopefully be consolidated once we have upgraded DropWizard from 1.0.2 to 1.1.0.

### DIFF
--- a/dependency-check-suppressions.xml
+++ b/dependency-check-suppressions.xml
@@ -14,4 +14,13 @@
         <gav regex="true">^org\.apache\.commons:commons-collections4:.*$</gav>
         <cve>CVE-2015-6420</cve>
     </suppress>
+
+    <suppress>
+        <notes><![CDATA[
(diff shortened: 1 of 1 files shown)


karlbaker02:
I did attempt this but wasn't able to get around it. I don't think that it's necessarily a bad thing, and hopefully this won't be needed when we upgrade DropWizard. Happy if you want to spend some time playing about with this, but probably not worth too much effort here.

PR #263, opened 2017-04-26
Upgrade DropWizard to 1.1.0

karlbaker02:
This PR upgrades DropWizard from version 1.0.2 to 1.1.0 and some of its related dependencies. Certain tests have been modified to remove setup statements that triggered a newly introduced `UnnecessaryStubbingException` in Mockito.

Vulnerabilities `CVE-2015-2808` and `CVE-2013-2566` have been suppressed, as DropWizard uses TLSv1.2 by default and will only include the RC4 cipher if it is explicitly specified in `config.yaml` under an `excludedCipherSuites` property (see https://github.com/dropwizard/dropwizard/blob/9a311914295a5cc81a59f6dc5981286a57b0284f/docs/source/manual/core.rst under heading `SSL`). All vulnerabilities in `tomcat-jdbc-8.5.9.jar` have been suppressed as we do not use Tomcat.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,12 +71,6 @@ dependencies {
     compile(postgresClient) {
         exclude group: 'org.slf4j'
     }
-    compile(logbackClassic){
-        force = true
-    }
-    compile(logbackAccess){
-        force = true
-    }
 
     testCompile junit, mockito, dropwizardTest, wiremock
     testCompile dropwizardTest, jsoup, jsonAssert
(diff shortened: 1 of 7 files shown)


samcrang:
I assume this isn't causing you occasional failures when running the whole test suite (like it did when I tried to do this upgrade ages ago)?

karlbaker02:
There is the occasional one test failing, which I'm not sure is what you're referring to. Let's collaborate.

karlbaker02:
Good spot John.

PR #264, opened 2017-04-27
Add script to spin up a register environment locally

samcrang:
Opinions?

### DIFF
--- a/config.docker.register.yaml
+++ b/config.docker.register.yaml
@@ -20,9 +20,8 @@ server:
 
 registerDomain: openregister.dev:8080
 
-register: school
-
-schema: school
+register: school-eng
+schema: school-eng
 
 enableDownloadResource: true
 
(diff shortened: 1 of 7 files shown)


samcrang:
The `./load-school-data.sh` step doesn't actually work—I'm looking at fixing it now.

samcrang:
This is fixed now.

At the moment it's a bit awkward for someone to access the basic registers in their browser without doing some `Host` header trickery. If we decide this is useful we can make this easier by adding some `A` records pointing to `127.0.0.1` for `{register,field,datatype}.local.openregister.org`.

config.docker.register.yaml
```diff
@@ -33,9 +32,8 @@ externalConfigDirectory: /tmp
 downloadConfigs: true
 
 # can be http: file s3: or classpath: url
-fieldsYamlLocation: classpath://config/fields.yaml
-
-registersYamlLocation: classpath://config/registers.yaml
+fieldsYamlLocation: http://field/records.yaml?page-size=5000
```
johnollier: If the ENVIRONMENT variable were passed in we should be able to allow the app to read direct from field.alpha.openregister.org etc In dropwizard we should be able to use field.${ENVIRONMENT:alpha}.openregister.org . Could we then dispense with docker-compose-basic?

samcrang: The point of this thing is that people can spin up and experiment with the whole stack--we want people to have these registers running locally. I'm only pre-loading them with data to make the initial steps less painful.

samcrang:
Also I didn't realise that `school-data.*` was used as part of the conformance tests 😩.

PR #269, opened 2017-05-03
Resolve indexes

johnollier:
There needs to be some configuration that maps the index names in the config file to IndexFunction classes in the project. I've tried to create this in a type safe way. The application will refuse to start if the index names do not match the names of the functions in the enumeration.


src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -207,9 +220,13 @@ public RegisterAuthenticator getAuthenticator() {
         return similarRegisters;
     }
 
+    public List<IndexFunctionConfiguration> getIndexFunctionConfigs() {
```
michaelabenyohai: This isn't used anywhere so can we remove it?

johnollier: It's removed.

johnollier:
I've changed the error message when no index function matches what's in the config.yaml.

PR #272, opened 2017-05-10
Redirect HTTP to HTTPS if application is behind a load balancer

samcrang:
If the application is running behind a load balancer and forwards the
original protocol the user used in the `X-Forwarded-Proto` header and
the original protocol is HTTP then upgrade to HTTPS.

### DIFF
--- a/src/main/java/uk/gov/register/filters/HttpToHttpsRedirectFilter.java
+++ b/src/main/java/uk/gov/register/filters/HttpToHttpsRedirectFilter.java
@@ -1,0 +1,44 @@
+package uk.gov.register.filters;
+
+import com.google.common.net.HttpHeaders;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+
+@Provider
+public class HttpToHttpsRedirectFilter implements javax.servlet.Filter {
+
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/register/filters/HttpToHttpsRedirectFilter.java
```diff
@@ -0,0 +1,44 @@
...
+
+   @Override
+   public void destroy() { }
+
+   @Override
+   public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+      HttpServletRequest httpRequest = (HttpServletRequest)request;
+
+      if ("http".equals(httpRequest.getHeader(HttpHeaders.X_FORWARDED_PROTO))) {
+         StringBuilder location = new StringBuilder();
```
johnollier: httpRequest.getRequestURL() returns the complete url except the query string as a StringBuffer - but that would only save a couple of lines.

samcrang: > Reconstructs the URL the client used to make the request. The returned URL contains a protocol, server name, port number, and server path, but it does not include query string parameters. 

I think we delibitely want to avoid pulling in the port number because if it's explicitly set to 80 (for some reason) we don't want to redirect to `https://` on the same port. Also we'd have to do a string replace to swap the scheme which feels a bit clunky.

PR #274, opened 2017-05-11
Make record(s) rendering consistent for indexes and registers

michaelabenyohai:
See individual commit messages for more detail.

This change implements more consistent rules for rendering `/records` and `/record/x` pages that work for both indexes and registers (no matter which is "configured"). The rules in detail are here: https://trello.com/c/N2Q7KGKi/784-make-rendering-of-record-x-and-records-more-consistent-for-registers-and-indexes.

I have attempted to tidy up the RecordView and RecordsViews as I feel that much of the logic belonged in RecordsView - in most formats, the two are now the same.


michaelabenyohai:
@johnollier me too, only just! Hopefully even more red coming soon.

PR #275, opened 2017-05-11
Introduce interface for data access

michaelabenyohai:
See individual commit messages for details. Overall, the changes are:

- Introduce an interface for data access - currently implements the same methods that exist now
- Put database and batching logic below the interface
- Keep only register domain logic above the interface
- Removed the concept of read/write domain objects
- Make performing a read from the RecordIndex force all entry and item data to be flushed to the database. This means that we could remove `register.commit()` hack from the `IndexDriver`.

This is a work in progress and the next steps include:

- remove concept of "on demand register" and "transactional register" - use the same db handle logic everywhere
- simplify and consolidate the number of methods in the DataAccessLayer interface.


karlbaker02:
This is very exciting! All the changes look good, definitely a vast improvement from what we currently have. If I get some spare time I'll test it out and update this PR if no-one beats me to it.

src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -89,49 +91,41 @@ public int migrate() {
...
+        return indexFunctionConfigs.stream().flatMap(c -> c.getIndexFunctions().stream()).collect(toList());
+    }
+
     public Register buildOnDemandRegister() {
         return new PostgresRegister(getRegisterMetadata(),
                 getRegisterFieldsConfiguration(),
-                new UnmodifiableEntryLog(memoizationStore.get(), dbi.onDemand(EntryQueryDAO.class), dbi.onDemand(IndexQueryDAO.class)),
-                new UnmodifiableItemStore(dbi.onDemand(ItemQueryDAO.class)),
-                new UnmodifiableRecordIndex(dbi.onDemand(RecordQueryDAO.class)),
+                new EntryLogImpl(getOnDemandDataAccessLayer(), memoizationStore.get()),
```
johnollier: Do we want to create >1 DataAccessLayer object at this point?

src/main/java/uk/gov/register/store/postgres/PostgresReadDataAccessLayer.java
```diff
@@ -0,0 +1,168 @@
...
+import uk.gov.register.core.Record;
+import uk.gov.register.db.*;
+import uk.gov.register.store.DataAccessLayer;
+import uk.gov.register.util.HashValue;
+
+import java.time.Instant;
+import java.util.*;
+import java.util.function.Function;
+
+public abstract class PostgresReadDataAccessLayer implements DataAccessLayer {
```
johnollier: As there's only one subclass of this, can it be combined with PostgresDataAccessLayer?

src/main/java/uk/gov/register/store/postgres/PostgresReadDataAccessLayer.java
```diff
@@ -0,0 +1,168 @@
...
+    public List<Record> getIndexRecords(int limit, int offset, String indexName) {
+        return indexQueryDAO.findRecords(limit, offset, indexName);
+    }
+
+    @Override
+    public int getTotalIndexRecords(String indexName) {
+        return indexQueryDAO.getTotalRecords(indexName);
+    }
+
+    protected void checkpoint() {
```
johnollier: Is this needed?

src/main/java/uk/gov/register/store/postgres/PostgresReadDataAccessLayer.java
karlbaker02: Wouldn't that render meaningless the distinction of having a separate read access layer from the write access layer?

karlbaker02: Wouldn't this render meaningless the distinction of having a separate read access layer from the write access layer?

michaelabenyohai: As discussed I kept these separate just to make it easier to visually see where the batching logic is and the straight-to-database logic is. In the work coming up I think these two classes will get a lot smaller and we should be able to combine them or split them in a more sensible way.

src/main/java/uk/gov/register/core/RegisterContext.java
michaelabenyohai: Yes! Fixed this.

src/main/java/uk/gov/register/store/postgres/PostgresReadDataAccessLayer.java
michaelabenyohai: This is needed by all the methods in this class. I have made it more obvious that it needs to be overridden by making it abstract. I think this will go away in the next iteration or two.

PR #296, opened 2017-06-22
Move current keys to index [DO NOT MERGE yet...]

karlbaker02:
This PR removes the usage of the `current_keys` table from DAO objects and instead uses indexes to store and retrieve the records of a register.

Joint venture with @samcrang

### DIFF
--- a/src/main/java/uk/gov/register/indexer/IndexDriver.java
+++ b/src/main/java/uk/gov/register/indexer/IndexDriver.java
@@ -17,7 +17,7 @@ public class IndexDriver {
     }
 
     public void indexEntry(Register register, Entry entry, IndexFunction indexFunction) {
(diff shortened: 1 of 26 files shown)


src/main/java/uk/gov/register/indexer/IndexDriver.java
```diff
@@ -17,16 +17,17 @@ public IndexDriver(DataAccessLayer dataAccessLayer) {
     }
 
     public void indexEntry(Register register, Entry entry, IndexFunction indexFunction) {
-        Optional<Record> currentRecord = register.getRecord(entry.getKey());
+        Optional<Record> currentRecord = register.getDerivationRecord(entry.getKey(), indexFunction.getName());
```
michaelabenyohai: This doesn't look correct. It should always get the record. i.e. it should always be getting the current record for the `records` index no matter what index is being written.

PR #307, opened 2017-07-25
Associate index functions with either user and system entry types

johnollier:
Previously the index functions would check the entry type. This change
associates the index function with an entry type so separate lists of
index functions can be applied as data is loaded.

### DIFF
--- a/src/main/java/uk/gov/register/core/PostgresRegister.java
+++ b/src/main/java/uk/gov/register/core/PostgresRegister.java
@@ -31,8 +31,7 @@ public class PostgresRegister implements Register {
     private final EntryLog entryLog;
     private final ItemStore itemStore;
     private final IndexDriver indexDriver;
-    private final List<IndexFunction> userIndexFunctions;
-    private final List<IndexFunction> systemIndexFunctions;
+    private final Map<EntryType,Collection<IndexFunction>> indexFunctionsByEntryType;
     private final ItemValidator itemValidator;
     private final EnvironmentValidator environmentValidator;
 
@@ -44,8 +43,7 @@ public class PostgresRegister implements Register {
                             ItemStore itemStore,
                             RecordIndex recordIndex,
                             DerivationRecordIndex derivationRecordIndex,
-                            List<IndexFunction> userIndexFunctions,
-                            List<IndexFunction> systemIndexFunctions,
+                            Map<EntryType,Collection<IndexFunction>> indexFunctionsByEntryType,
                             IndexDriver indexDriver,
                             ItemValidator itemValidator,
                             EnvironmentValidator environmentValidator) {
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -91,8 +91,14 @@ public int migrate() {
         return flyway.migrate();
     }
 
-    private List<IndexFunction> getIndexFunctions() {
-        return indexFunctionConfigs.stream().flatMap(c -> c.getIndexFunctions().stream()).collect(toList());
+    private List<IndexFunction> getUserIndexFunctions() {
```
karlbaker02: Could we maybe just have a method which instead returns a `Map<EntryType, List<IndexFunction>`, which then removes the need for us to put an additional `if` statement inside of `PostgresRegister` and means that we only need the one method here?

johnollier:
I've made some changes in line with Karl's comments.

src/main/java/uk/gov/register/indexer/function/AddAllIndexFunction.java
```diff
@@ -7,15 +7,13 @@
 
 import java.util.Set;
 
-public class MetadataIndexFunction extends BaseIndexFunction {
-    public MetadataIndexFunction(String name) {
+public class AddAllIndexFunction extends BaseIndexFunction {
```
karlbaker02: Perhaps we rename this to something like `LatestByKey`, given that the index function isn't controlling what is and isn't being appended to the register?

src/main/java/uk/gov/register/core/PostgresRegister.java
```diff
@@ -67,30 +67,33 @@ public void putItem(Item item) {
...
-        for (IndexFunction indexFunction : indexFunctions) {
-            indexDriver.indexEntry(this, entry, indexFunction);
-        }
-
         if (entry.getEntryType() == EntryType.user) {
+            for (IndexFunction indexFunction : indexFunctionsByEntryType.get(EntryType.user)) {
+                indexDriver.indexEntry(this, entry, indexFunction);
+            }
             recordIndex.updateRecordIndex(entry);
+        } else {
```
karlbaker02: We can remove this else and do the indexing as we know the entry type, then update the record index only if it's a user entry

src/main/java/uk/gov/register/indexer/function/AddAllIndexFunction.java
michaelabenyohai: 👍

src/main/java/uk/gov/register/core/PostgresRegister.java
michaelabenyohai: 👍

PR #308, opened 2017-07-26
Move current keys to index

karlbaker02:
This PR removes the usage of the current_keys table from DAO objects and instead uses indexes to store and retrieve the records of a register. It includes a flyway migration to ensure that any existing records are created in the index table for current registers.


src/main/java/uk/gov/register/configuration/IndexFunctionConfiguration.java
```diff
@@ -55,6 +53,7 @@ public String getName() {
     public static class IndexNames {
         public static final String CURRENT_COUNTRIES = "current-countries";
         public static final String LOCAL_AUTHORITY_BY_TYPE = "local-authority-by-type";
+        public static final String RECORDS = "records";
```
michaelabenyohai: I'm wondering whether we should call this "record" rather than "records"?

karlbaker02: Perhaps so

src/main/java/uk/gov/register/store/DataAccessLayer.java
```diff
@@ -44,9 +44,9 @@
     Optional<Record> getRecord(String key);
     List<Record> getRecords(int limit, int offset);
     List<Record> findMax100RecordsByKeyValue(String key, String value);
-    Collection<Entry> findAllEntriesOfRecordBy(String key);
-
-    int getTotalRecords();
+    Collection<Entry> getAllEntriesByKey(String key);
+    //int getTotalRecords();
```
michaelabenyohai: Can this be deleted?

src/main/java/uk/gov/migration/R__11_Migrate_current_keys_to_index.java
```diff
@@ -0,0 +1,220 @@
...
+		TempDataSource ds = new TempDataSource(connection);
+		DBI dbi = new DBI(ds);
+		dbi.registerContainerFactory(new OptionalContainerFactory());
+		Handle handle = dbi.open();
+		
+		DataAccessLayer dataAccessLayer = new PostgresDataAccessLayer(
+				handle.attach(EntryQueryDAO.class),
+				handle.attach(IndexDAO.class),
+				handle.attach(IndexQueryDAO.class),
+				null,
```
michaelabenyohai: Is it possible to not make these `null`? Seems a bit nasty to leave them as `null`.

PR #310, opened 2017-07-26
Explicitly specify the tag of the Docker image

samcrang:
This means that we can reason about which version of images are running.
We know that it'll be a specific (hopefully immutable) tag and not just
what happens to be "latest" at a given point. This also makes updating
images less tricky—you just update the tag, do a deploy and the new
image gets pulled in.

### DIFF
--- a/deploy/scripts/start-service.sh
+++ b/deploy/scripts/start-service.sh
@@ -27,7 +27,7 @@ docker run \
     --network openregisters \
     --env REGISTER_NAME=${REGISTER_NAME} \
     --env DATACENTER=${DATACENTER} \
-    telegraf
+    telegraf:1.3.4
 
 docker run \
     --detach \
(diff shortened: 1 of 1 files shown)


deploy/scripts/start-service.sh
```diff
@@ -49,7 +49,7 @@ docker run \
     --network openregisters \
     --log-driver=fluentd \
     --log-opt fluentd-async-connect=true \
-    openjdk:8-jre-alpine \
```
philandstuff: have you deliberately chosen to move away from the alpine-based image? `8u131-jre-alpine` is available

samcrang: I put something in the git commit that introduced this:

> Upgrading to the latest version (8u141) as well as moving to using
images based off Debian base images. It seems that Debian have more
timely responses to security issues with packages. Version 8u141 has
been available for >7 days, has been in Debian Stretch >2 days but still
hasn't landed in a release version of Alpine. It has been available in
Alpine Edge for >7 days but not in an actual release.

philandstuff: sweet m8, sweet 👌

PR #312, opened 2017-08-01
Remove need for schema creation permissions when deploying to PaaS

samcrang:


src/main/resources/paas.yaml
```diff
@@ -26,8 +26,8 @@ externalConfigDirectory: /tmp
 
 downloadConfigs: true
 
-fieldsYamlLocation: classpath://config/fields.yaml
-registersYamlLocation: classpath://config/registers.yaml
+fieldsYamlLocation: https://field.discovery.openregister.org/records.yaml
```
karlbaker02: I'm guessing you didn't mean to include these changes here?

samcrang: I did. Do you not feel it's the right thing to do?

samcrang: Actually you're right—this should probably be configurable in `manifest.yaml` but with a sensible default.

karlbaker02: I was thinking more along the lines that we allow the phase to be specified (via register domain right?), so this probably shouldn't be pointing directly at discovery.

samcrang: Or just make it possible to set anything. Then you could have your own register/field registers deployed on PaaS.

samcrang:
How's this @karlbaker02?

PR #314, opened 2017-08-02
Add role to button link

### LINKED ISSUE #311: Buttons should have role=button so they behave like buttons
If you visit https://country.register.gov.uk/ using assistive tech like a screen reader the button will be read out as a link, and pressing space will not activate the button.

You'll need to add the attribute 'role=button' and the following JavaScript shim (which should be avaliable if you're using govuk_frontend_toolkit): 

https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/shim-links-with-button-role.js
https://github.com/alphagov/static/commit/3c9bc1bfa45537262c4c079f7d3219a32f96f121

danmitchell-:
Fixes #311


src/main/resources/assets/js/shim-links-with-button-role.js
```diff
@@ -0,0 +1,34 @@
...
+// Created since some Assistive Technologies (for example some Screenreaders)
+// Will tell a user to press space on a 'button', so this functionality needs to be shimmed
+// See https://github.com/alphagov/govuk_elements/pull/272#issuecomment-233028270
+//
+// Usage instructions:
+// GOVUK.shimLinksWithButtonRole.init();
+;(function (global) {
+  'use strict'
+
+  var $ = global.jQuery
```
samcrang: It doesn't look like we are using jQuery so `$` ends up being `undefined`

src/main/resources/assets/js/shim-links-with-button-role.js
```diff
@@ -0,0 +1,34 @@
...
+  var GOVUK = global.GOVUK || {}
+
+  GOVUK.shimLinksWithButtonRole = {
+
+    init: function init () {
```
samcrang: We doesn't appear that this function is invoked by anything?

danmitchell-:
It's massively overkill to include jQuery for just the button shim but it's what we do on govuk_elements. :man_shrugging:

danmitchell-:
jQuery version is taken from [govuk_elements](https://github.com/alphagov/govuk_elements/blob/master/assets/javascripts/vendor/jquery-1.11.0.min.js)

nickcolley:
Thanks for sorting this so quickly! 🚀 

The jQuery dep is sad indeed, maybe our accessibility shims should be library agnostic... @danmitchell- worth raising on govuk elements?

I think this'd also allow us to comfortably [cut the mustard](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard) without leaving older browsers behind that rely on shims like this.

PR #315, opened 2017-08-07
Move records to indexes and improve performance

karlbaker02:
This PR removes the usage of the current_keys table from DAO objects and instead uses indexes to store and retrieve the records of a register. It includes a flyway migration to ensure that any existing records are created in the index table for current registers.

It also makes several refactors to improve performance when loading large amounts of data, removing the need to continually call the database by keeping the same equivalent data in memory.


src/main/java/uk/gov/register/indexer/function/RecordIndexFunction.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.register.indexer.function;
```
michaelabenyohai: Can we remove this? I think we're using `LatestByKeyIndexFunction.java` in place of this now. This is the same for `MetadataIndexFunction.java`, which seems to exist as an empty file - can we remove it?

src/main/java/uk/gov/register/resources/RecordResource.java
```diff
@@ -45,7 +50,7 @@ public RecordResource(RegisterReadOnly register, ViewFactory viewFactory, Reques
     public RecordView getRecordByKey(@PathParam("record-key") String key) {
         httpServletResponseAdapter.addLinkHeader("version-history", String.format("/record/%s/entries", key));
 
-        return register.getRecord(key).map(viewFactory::getRecordMediaView)
+        return register.getDerivationRecord(key, IndexNames.RECORD).map(viewFactory::getRecordMediaView)
```
michaelabenyohai: I don't believe this change needs to be made. The `getRecord()` method in `PostgresRegister` maps down to `getDerivationRecord(key, IndexNames.RECORD)` as well.

src/main/java/uk/gov/register/resources/RecordResource.java
```diff
@@ -118,7 +123,7 @@ public RecordsView records(@QueryParam(IndexSizePagination.INDEX_PARAM) Optional
     }
 
     private IndexSizePagination setUpPagination(@QueryParam(IndexSizePagination.INDEX_PARAM) Optional<IntegerParam> pageIndex, @QueryParam(IndexSizePagination.SIZE_PARAM) Optional<IntegerParam> pageSize) {
-        IndexSizePagination pagination = new IndexSizePagination(pageIndex.map(IntParam::get), pageSize.map(IntParam::get), register.getTotalRecords());
+        IndexSizePagination pagination = new IndexSizePagination(pageIndex.map(IntParam::get), pageSize.map(IntParam::get), register.getTotalDerivationRecords(IndexNames.RECORD));
```
michaelabenyohai: Similar to the above. We need to make these two things consistent - do we do the mapping to the correct index function in `PostgresRegister` or here. I know we remove `getTotalRecords()` from the `PostgresRegister` but maybe this could be considered a common function that needs to exist on the `Register` interface.

src/main/java/uk/gov/register/indexer/IndexDriver.java
```diff
@@ -1,48 +1,59 @@
...
-        this.dataAccessLayer = dataAccessLayer;
+    public IndexDriver() {
+        this.currentIndexEntryNumbers = new HashMap<>();
     }
 
-    public void indexEntry(RegisterReadOnly register, Entry entry, IndexFunction indexFunction) {
-        Optional<Record> currentRecord = (entry.getEntryType() == EntryType.user)
-                ? register.getRecord(entry.getKey())
-                : register.getDerivationRecord(entry.getKey(), IndexNames.METADATA);
+    public void indexEntry(DataAccessLayer dataAccessLayer, Entry entry, IndexFunction indexFunction,
```
michaelabenyohai: It might be better to keep `DataAccessLayer` as part of the constructor rather than as an argument to this method?

src/main/java/uk/gov/migration/R__11_Migrate_current_keys_to_index.java
```diff
@@ -32,46 +32,56 @@ public void migrate(Connection connection) throws Exception {
...
+				connection.getSchema(),
+				new IndexDriver(),
+				ImmutableMap.of(EntryType.user, Arrays.asList(new RecordIndexFunction(IndexNames.RECORD))));
 
 		RegisterReadOnly register = new RecordIndexRegister(dataAccessLayer);
-		IndexDriver indexDriver = new IndexDriver(dataAccessLayer);
+		IndexDriver indexDriver = new IndexDriver();
+
+		int recordCount = register.getTotalDerivationRecords(IndexFunctionConfiguration.IndexNames.RECORD);
+		Map<String, Record> indexRecords = register.getDerivationRecords(recordCount, 0, IndexNames.RECORD).stream().collect(Collectors.toMap(k -> k.getEntry().getKey(), v -> v));
```
michaelabenyohai: Can we get this information from `dataAccessLayer`? Then we wouldn't need the `RecordIndexRegister` private class at all.

src/main/java/uk/gov/register/serialization/RSFExecutor.java
```diff
@@ -71,9 +71,14 @@ private RegisterResult validate(RegisterCommand command, Register register, int
         if (commandName.equals("add-item")) {
             validateAddItem(command, rsfLine, hashRefLine);
         } else if (commandName.equals("append-entry")) {
+            // Check that we should still allow system entries
+            if (allowSystemEntries) {
+                allowSystemEntries = register.getTotalStagedEntries(EntryType.user) == 0 ? register.getTotalEntries(EntryType.user) == 0 : false;
```
michaelabenyohai: Can we remove the need to call `getTotalStagedEntries(EntryType.user)` here? Could we change the `getTotalEntries(EntryType entryType)` method in `PostgresDataAccessLayer` to read from `stagedEntries` if it needs to?

src/main/java/uk/gov/register/store/DataAccessLayer.java
```diff
@@ -22,35 +19,38 @@
...
     Collection<Entry> getAllEntries();
+
     Iterator<Entry> getEntryIterator();
     Iterator<Entry> getEntryIterator(int totalEntries1, int totalEntries2);
     Iterator<Entry> getIndexEntryIterator(String indexName);
     Iterator<Entry> getIndexEntryIterator(String indexName, int totalEntries1, int totalEntries2);
     <R> R withEntryIterator(Function<EntryIterator, R> callback);
     int getTotalEntries();
     int getTotalEntries(EntryType entryType);
+    int getTotalStagedEntries(EntryType entryType);
```
michaelabenyohai: Can we remove the need for this? It would be good to keep the knowledge of staged data below this interface.

src/main/java/uk/gov/register/store/DataAccessLayer.java
```diff
@@ -22,35 +19,38 @@
...
     Collection<Item> getAllItems();
+
     Iterator<Item> getItemIterator();
     Iterator<Item> getItemIterator(int start, int end);
     Iterator<Item> getSystemItemIterator();
     
     // Record Index
-    void updateRecordIndex(Entry entry);
     Optional<Record> getRecord(String key);
+    List<Record> getRecords(List<String> keys);
```
michaelabenyohai: Is this used anywhere? Can we remove it?

karlbaker02: Removed from PostgresDataAccessLayer too.

src/main/java/uk/gov/register/store/postgres/PostgresReadDataAccessLayer.java
```diff
@@ -153,16 +158,20 @@ public int getTotalSystemEntries() {
     }
 
     // Index
-
     @Override
     public Optional<Record> getIndexRecord(String key, String indexName) {
-        checkpoint();
-        return indexQueryDAO.findRecord(key, indexName, schema, indexName.equals(IndexNames.METADATA) ? "entry_system" : "entry");
+        List<Record> records = indexQueryDAO.findRecords(Arrays.asList(key), indexName, schema, indexName.equals(IndexNames.METADATA) ? "entry_system" : "entry");
```
michaelabenyohai: Could use `getIndexRecords(Arrays.asList(key))` here, which de-duplicates some logic.

src/main/resources/sql/V10__Index_on_index_table.sql
```diff
@@ -0,0 +1,7 @@
+create index if not exists index_name on index (name);
+create index if not exists index_key on index (key);
+create index if not exists index_sha256hex on index (sha256hex);
+create index if not exists index_start_entry_number on index (start_entry_number);
+create index if not exists index_end_entry_number on index (end_entry_number);
+create index if not exists index_start_index_entry_number on index (start_index_entry_number);
+create index if not exists index_end_index_entry_number on index (end_index_entry_number);
```
michaelabenyohai: There are a lot of indexes here and all on individual columns. Do these indexes help with loading and/or querying? Do we still need them if we also have batching?

karlbaker02: When I added the indexes on all these, we were doing lookups on each of these columns from what I saw over all of our different queries. Happy to revisit this again.

src/main/java/uk/gov/register/store/postgres/PostgresDataAccessLayer.java
```diff
@@ -14,38 +17,79 @@
...
         stagedEntries.add(entry);
+        stagedEntryKeys.add(entry.getKey());
 
         if (entry.getEntryType().equals(EntryType.system)) {
             checkpoint();
         }
     }
+    
+    @Override
+    public int getCurrentIndexEntryNumber(String indexName) {
```
michaelabenyohai: This doesn't have any batching specific logic in it and the method hasn't changed - should we leave it in `PostgresReadDataAccessLayer.java` where it was previously?

src/main/java/uk/gov/register/store/DataAccessLayer.java
```diff
@@ -22,35 +19,38 @@
...
     List<Record> findMax100RecordsByKeyValue(String key, String value);
     Collection<Entry> getAllEntriesByKey(String key);
-    //int getTotalRecords();
     
     // Index
     void start(String indexName, String key, String itemHash, int startEntryNumber, int startIndexEntryNumber);
-    void end(String indexName, String entryKey, String indexKey, String itemHash, int endEntryNumber, int endIndexEntryNumber);
+    void end(String indexName, String entryKey, String indexKey, String itemHash, int endEntryNumber, int endIndexEntryNumber, int entryNumberToEnd);
     Optional<Record> getIndexRecord(String key, String indexName);
+    List<Record> getIndexRecords(List<String> keys, String indexName);
```
michaelabenyohai: Do we need this? This method only appears to be used below the `DataAccessLayer` interface.

src/main/resources/sql/V10__Index_on_index_table.sql
michaelabenyohai: Are most of the look ups on combined sets of these columns? If so, the index would need to be across multiple columns? I think we need to revisit this. Adding an index to ever individual column could really slow down loads when the database gets large.

src/main/java/uk/gov/register/store/postgres/PostgresDataAccessLayer.java
```diff
@@ -14,38 +17,79 @@
...
+        return indexQueryDAO.getCurrentIndexEntryNumber(indexName, schema);
+    }
+
+    @Override
+    public IndexEntryNumberItemCountPair getStartIndexEntryNumberAndExistingItemCount(String indexName, String key, String sha256hex){
```
michaelabenyohai: This method doesn't account for the fact that the index table could have some rows in it that need to be included in this query. I think this might be a problem?

src/main/java/uk/gov/register/indexer/IndexDriver.java
michaelabenyohai: @karlbaker02 what do you think about this?

src/main/java/uk/gov/migration/R__11_Migrate_current_keys_to_index.java
michaelabenyohai: @karlbaker02 as discussed, would still be good to do this.

karlbaker02: Done.

src/main/resources/sql/V10__Index_on_index_table.sql
karlbaker02: This migration has been removed.

PR #316, opened 2017-08-07
perform required submodule steps in ./go script

gidsg:
...turns out without these the `./gradlew clean build` script fails as follows:

```
Errno::ENOENT on line ["794"] of org/jruby/RubyFile.java: No such file or directory - /Users/gideongoldberg/git/openregister-java/src/main/resources/assets/govuk_frontend_toolkit/stylesheets
Run with --trace to see the full backtrace
:compassCompile FAILED
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ If the basic registers change you can restart your register so that it sees the 
 - Install and run postgres db
 - Checkout project
 - Run the `./go` script to set everything up
+- `$ git submodule init`
+- `$ git submodule update`
 - Build project using command `./gradlew clean build`
 
 ### Working on the frontend
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ If the basic registers change you can restart your register so that it sees the 
 - Install and run postgres db
 - Checkout project
 - Run the `./go` script to set everything up
-- `$ git submodule init`
 - `$ git submodule update`
 - Build project using command `./gradlew clean build`
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ If the basic registers change you can restart your register so that it sees the 
 - Install and run postgres db
 - Checkout project
 - Run the `./go` script to set everything up
-- `$ git submodule update`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,6 +48,8 @@ If the basic registers change you can restart your register so that it sees the
 - Install and run postgres db
 - Checkout project
 - Run the `./go` script to set everything up
+- `$ git submodule init`
+- `$ git submodule update`
```
samcrang: Looks like `git submodule init` is already in `go`. Does it make sense for `git submodule update` to go in that script too?

karlbaker02: It needs to be done when you first clone the repo, so possibly.

danmitchell-: I don't think we want to have `git submodule update` in the `go` script since this will update the submodules which could have adverse effects to the UI

gidsg: So, erm, I should probably remove the `init` step since that sounds it's being duplicated and just leave `update`? @karlbaker02

karlbaker02: @danmitchell- Are you saying that the `git submodule update` should be run manually by a user and not as part of the `go` script in case they do not realise that the `go` script is doing this update?

README.md
```diff
@@ -48,7 +48,6 @@ If the basic registers change you can restart your register so that it sees the
 - Install and run postgres db
 - Checkout project
 - Run the `./go` script to set everything up
-- `$ git submodule init`
 - `$ git submodule update`
```
karlbaker02: Let's remove this line and instead edit the `go` script to replace:

`git submodule init`

with 

`git submodule update --init`

Which will 'update' submodules to the expected version and initialize any that haven't yet been initialized.

README.md
danmitchell-: After further research we can replace `git submodule init` with `git submodule update --init`, this will pull down the submodules based on the commit version within the project and initialise if they aren't already. We need to add further documentation to actually update the submodules which is `git submodule foreach git pull origin master`

PR #317, opened 2017-08-07
Add IntelliJ run steps

gidsg:
...as this did not work out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ If the basic registers change you can restart your register so that it sees the 
 - `Run > Edit Configurations`
 - Edit `RegisterRunner`
 - Use classpath of module: `openregister-java_main`
-- JRE: `1.8`
+- JRE: *select your installed JRE e.g.* `1.8`
 - `Run > Run... > RegisterRunner`
 
 To test if everything has worked:
 
-`$ curl 127.0.0.1:8080/records -H "Host: field"` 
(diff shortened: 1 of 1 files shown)


karlbaker02:
This is cool. Just a few small comments I'll put in the appropriate places now...

README.md
```diff
@@ -50,6 +50,22 @@ If the basic registers change you can restart your register so that it sees the
 - Run the `./go` script to set everything up
 - Build project using command `./gradlew clean build`
 
+### IntelliJ setup
+- `File > Open` your checkout directory
+- `Run > Edit Configurations`
+- Edit `RegisterRunner`
+- Use classpath of module: `openregister-java_main`
+- JRE: `1.8`
```
karlbaker02: This will be whichever JRE the user is running, although probably 1.8

README.md
```diff
@@ -50,6 +50,22 @@ If the basic registers change you can restart your register so that it sees the
...
+- `Run > Run... > RegisterRunner`
+
+To test if everything has worked:
+
+`$ curl 127.0.0.1:8080/records -H "Host: field"`
```
karlbaker02: I'd maybe change the host header here to be `-H "Host: school"` as this is the register which will run on port `8080` when the register starts up

gidsg: When I do this it still returns `Register undefined` is that expected?

karlbaker02: Yep, that's expected.

PR #320, opened 2017-08-08
Read app configuration from S3 when running on PaaS

samcrang:
This allows us to generate our slightly complex configurations
separately and have them consumed by something running on PaaS without
having to bundle it into our `jar`.


src/main/resources/paas.yaml
```diff
@@ -1,42 +0,0 @@
-database:
```
gidsg: I wonder if it is worth keeping this file in git just for versioning reasons. If you disagree perhaps enable versioning on the file in S3 so there is a means to track changes?

samcrang: There's lots of examples of configuration files of a similar shape (although they do different things) already in this repo. I don't think there's much need to keep this around.

PR #326, opened 2017-08-16
Use CodeBuild to deploy to PaaS

samcrang:
This is very much a (mis)use of CodeBuild.

We currently deploy our code as part of a CodePipeline using CodeDeploy.
CodeDeploy works well with AWS-y things but doesn't have a way a neat
way to deploy to something externally like PaaS.

We can use CodeBuild to run a bunch of `cf` commands to push the `.jar`
to PaaS.

In order to do this our input to the CodeBuild project (the `.zip` that
Travis builds) has to contain a `buildspec.yml` and manifests the
describe the PaaS deployment--this is not ideal and our build artifact
has to know too much about its own deployment.

This feels very much like a stopgap solution to deploying to PaaS and
should be replaced if something more appropriate materializes.

There is a lot of duplication across manifests for different
environments. In the future I expect these manifests to define things
like the routes and domains associated with the app. Whilst they appear
similar now I think this is only accidental.

In order for this to work additional changes are required to
CodePipeline but the artifact must change first.

### DIFF
--- a/deploy/buildspec.yml
+++ b/deploy/buildspec.yml
@@ -1,0 +1,28 @@
+version: 0.2
+
+phases:
+  install:
+    commands:
+      - apt-get update -y
+      - apt-get install -y apt-transport-https
+      - wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | apt-key add -
+      - echo 'deb http://packages.cloudfoundry.org/debian stable main' > /etc/apt/sources.list.d/cloudfoundry-cli.list
+      - apt-get update -y
+      - apt-get install -y cf-cli
+  pre_build:
+    commands:
+      - export CF_USER="$(aws ssm get-parameters --name paas-deploy-user --query 'Parameters[0].Value' --output text)"
+      - export CF_PASSWORD="$(aws ssm get-parameters --name paas-deploy-password --with-decryption --query 'Parameters[0].Value' --output text)"
+      - export AWS_KEY="$(aws ssm get-parameters --name paas-deploy-aws-key --query 'Parameters[0].Value' --output text)"
+      - export AWS_SECRET="$(aws ssm get-parameters --name paas-deploy-aws-secret --with-decryption --query 'Parameters[0].Value' --output text)"
+      - cf api "https://api.cloud.service.gov.uk"
+      - cf auth "$CF_USER" "$CF_PASSWORD"
+      - cf target -o "$CF_ORGANIZATION" -s "$CF_SPACE"
+  build:
+    commands:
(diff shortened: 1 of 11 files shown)


deploy/buildspec.yml
```diff
@@ -0,0 +1,28 @@
...
+      - apt-get install -y apt-transport-https
+      - wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | apt-key add -
+      - echo 'deb http://packages.cloudfoundry.org/debian stable main' > /etc/apt/sources.list.d/cloudfoundry-cli.list
+      - apt-get update -y
+      - apt-get install -y cf-cli
+  pre_build:
+    commands:
+      - export CF_USER="$(aws ssm get-parameters --name paas-deploy-user --query 'Parameters[0].Value' --output text)"
+      - export CF_PASSWORD="$(aws ssm get-parameters --name paas-deploy-password --with-decryption --query 'Parameters[0].Value' --output text)"
+      - export AWS_KEY="$(aws ssm get-parameters --name paas-deploy-aws-key --query 'Parameters[0].Value' --output text)"
```
karlbaker02: Small question: do we not encrypt the aws key?

deploy/manifests/alpha/basic.yml
```diff
@@ -0,0 +1,15 @@
+---
+applications:
+- name: alpha-basic
+  memory: 1024M
+  instances: 2
+  buildpack: java_buildpack
+  health-check-http-endpoint: /records
+  random-route: true
+  services:
+    - alpha-db
```
karlbaker02: We currently have our DB connection strings defined in `config.yaml` - we will be keeping this I imagine and updating it to the new connection strings provided by PaaS? I imagine that here we're only saying that this particular application has access to this particular database?

samcrang: The connection strings are ephemeral and can change on bind/unbind/probably other things. We can't hardcode the them into the `config.yaml`, [so we have to do read them from the environment at runtime](https://github.com/openregister/openregister-java/blob/master/src/main/java/uk/gov/register/configuration/DatabaseManager.java#L22). We do however have to [provide a dummy value so Dropwizard doesn't barf](https://github.com/openregister/deployment/blob/master/ansible/roles/openregister_config/templates/openregister-config.yaml.j2#L17).

You're also right that we're describing that this app is even able to connect to this database—I believe there's some Postgres configuration that expects to only see connections from certain IPs.

deploy/buildspec.yml
samcrang: We do not encrypt the key. We do encrypt the secret though. I think this makes things easier to reason about if you were trying to work out who and what were using keys.

PR #327, opened 2017-08-17
Fix deprecation warning; use Python 3.6

gidsg:
Previously the build would display a warning as [`pyvenv` has been deprecated](https://docs.python.org/3.6/whatsnew/3.6.html#id8) 

This fixes the warning and switches to using Python 3.6 (which is currently the default when you do `brew install python3` by using the new built in `venv` functionality.


gidsg:
This is failing because we don't have Python 3.6 on Travis CI, I will take a look if that can be done.

PR #329, opened 2017-08-21
Default to cloning a smaller register when running locally

michaelabenyohai:
Previously we cloned the alpha school-eng register when using the `./run-application.sh` script. This is now a reasonably large register that takes a while to load and may also move to the beta environment soon. This change therefore defaults to cloning the beta country register on `./run-application.sh` as this is a smaller and more stable register.

This also adds new documentation on changing the default to clone a different existing register.

The existing documentation for creating a new register locally has been updated to account for metadata changes.

### DIFF
--- a/RUNNING_LOCALLY.md
+++ b/RUNNING_LOCALLY.md
@@ -38,6 +38,20 @@ If you want to amend this register, you can:
 * create a new register
 * change the configuration
 
+## Clone a different register
+
+The `./run-application.sh` script is configured to spin up a clone of the [Country beta register](https://country.register.gov.uk/). To clone a different register from another phase, you need to change some configuration.
+
+For example, you might want to clone the [School type register](https://school-type.alpha.openregister.org/) from the alpha phase.
+
+Go to the `config.docker.register.yaml` file.
+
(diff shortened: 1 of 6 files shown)


karlbaker02:
This looks good to me, all works locally too. Will let Gideon confirm once it's up and running on his machine.

RUNNING_LOCALLY.md
```diff
@@ -30,55 +30,82 @@ The command will create Docker containers that will:
...
+You should see your new undefined register at `127.0.0.1:8080`.
 
 ## Load data into your register
 
 You can now load your own data in the new register.
 
+First, you need to populate the register with its register definition and field definitions. These should be the same as those listed in the field and register registers.
+
+```
+$ echo '[{"name":"foobar"}]' | python3 ./scripts/json-to-rsf/json2rsf.py system name | curl 127.0.0.1:8080/load-rsf -H "Content-Type: application/uk-gov-rsf" --data-binary @- -u foo:bar
```
gidsg: This is quite a lot of commands to run. Perhaps consider chaining them with `&&` ?

PR #330, opened 2017-08-21
Make deploys zero-downtime

samcrang:
We lean on the `cf-blue-green-deploy`[1] to make this as easy as
possible.

[1] https://github.com/bluemixgaragelondon/cf-blue-green-deploy

### DIFF
--- a/deploy/buildspec.yml
+++ b/deploy/buildspec.yml
@@ -9,6 +9,13 @@ phases:
       - echo 'deb http://packages.cloudfoundry.org/debian stable main' > /etc/apt/sources.list.d/cloudfoundry-cli.list
       - apt-get update -y
       - apt-get install -y cf-cli
+      - export CF_BGD_VERSION="1.2.0"
+      - export CF_BGD_CHECKSUM="a5677bf250077858dc2d39af12f48803a7d852a400efa246887556247e1203c4"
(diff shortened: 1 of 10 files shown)


karlbaker02:
With our current deployment in CodeDeploy we deploy to one instance at a time, and wait for the `validate.sh` script to complete before moving on to the next deploy. Do we still do this with PaaS? Regarding the blue-green plugin, we're not passing any tests to run so the only reason we're using it is to automate the process of deploying, swapping out the old for the new and cleaning the old?

samcrang:
On `validate.sh`: the `manifest.yaml` specifies a `healthcheck: /records`—I'd expect this to cause the deploy to fail and drop out but I still need to verify what actually happens.

And yes, the blue-green plugin [eliminates the need to do something like this](https://docs.cloudfoundry.org/devguide/deploy-apps/blue-green.html).

And you're right we have no smoke tests at the moment—we should probably create some.

samcrang:
@karlbaker02 I've run through a couple of failure scenarios and discovered the health check wasn't actually working—I've fixed that in the manifest and failed deploys work like you'd expect (ie. the deploy stops, is deteceted as a failure and the original app still recieves all the traffic).

karlbaker02:
Discussed above issues with Sam.

PR #331, opened 2017-08-22
Add routes

samcrang:
- Add routes so PaaS responds to requests with the correct app.
- Explicitly specify the syslog -> logit drain.
- Address doesn't need to run on its own instances anymore.


karlbaker02:
You may need to tinker with the routes between now and when we deploy to PaaS if any of them are removed / move phases.

PR #338, opened 2017-09-05
Validate datetime server side: Adding ISO_8601 format valitation for …

fjavierm:
…datetime type following especification located http://openregister.github.io/specification/#datetime-datatype

### DIFF
--- a/src/main/java/uk/gov/register/service/ItemValidator.java
+++ b/src/main/java/uk/gov/register/service/ItemValidator.java
@@ -3,7 +3,6 @@ package uk.gov.register.service;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.Sets;
 import org.apache.commons.lang3.StringUtils;
-import uk.gov.register.configuration.ConfigManager;
 import uk.gov.register.core.Cardinality;
 import uk.gov.register.core.Field;
 import uk.gov.register.core.RegisterMetadata;
--- a/src/main/java/uk/gov/register/util/DateFormat.java
+++ b/src/main/java/uk/gov/register/util/DateFormat.java
@@ -1,0 +1,14 @@
+package uk.gov.register.util;
+
+public interface DateFormat {
+
+    interface ISO_8601 {
+        String LABEL = "ISO_8601";
+
+        String YEAR = "yyyy";
+        String YEAR_MONTH = "yyyy-MM";
+        String YEAR_MONTH_DAY = "yyyy-MM-dd";
+        String YEAR_MONTH_DAY_HOURS_MINUTES_SECONDS = "yyyy-MM-dd'T'hh:mm:ss";
+        String YEAR_MONTH_DAY_HOURS_MINUTES_SECONDS_UTC = "yyyy-MM-dd'T'hh:mm:ss'Z'";
+    }
+}
--- a/src/main/java/uk/gov/register/util/DateFormatChecker.java
+++ b/src/main/java/uk/gov/register/util/DateFormatChecker.java
@@ -1,0 +1,25 @@
+package uk.gov.register.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.time.DateUtils;
+
+import java.text.ParseException;
+
+import static uk.gov.register.util.DateFormat.ISO_8601.*;
+
+public class DateFormatChecker {
+
+    public static boolean isDateFormatValid(final String dateToValidate) {
+        if (StringUtils.isEmpty(dateToValidate)) {
+            return true;
+        }
+
+        try {
+            DateUtils.parseDate(dateToValidate, YEAR, YEAR_MONTH, YEAR_MONTH_DAY, YEAR_MONTH_DAY_HOURS_MINUTES_SECONDS, YEAR_MONTH_DAY_HOURS_MINUTES_SECONDS_UTC);
+
+            return true;
+        } catch (ParseException e) {
+            return false;
+        }
+    }
+}
--- a/src/test/java/uk/gov/register/service/ItemValidatorTest.java
+++ b/src/test/java/uk/gov/register/service/ItemValidatorTest.java
@@ -28,9 +28,14 @@ public class ItemValidatorTest {
     @Mock
     private RegisterMetadata registerMetadata;
 
+    @Mock
+    private RegisterMetadata countryRegisterMetadata;
+
     private Map<String,Field> fieldsByName;
+    private Map<String,Field> countryFieldsByName;
 
     private ItemValidator itemValidator;
+    private ItemValidator countryItemValidator;
 
     private final ObjectMapper objectMapper = new ObjectMapper();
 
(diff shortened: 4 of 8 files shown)


src/main/java/uk/gov/register/util/DateFormat.java
```diff
@@ -0,0 +1,14 @@
...
+
+public interface DateFormat {
+
+    interface ISO_8601 {
+        String LABEL = "ISO_8601";
+
+        String YEAR = "yyyy";
+        String YEAR_MONTH = "yyyy-MM";
+        String YEAR_MONTH_DAY = "yyyy-MM-dd";
+        String YEARS_MONTH_DATA_TIME = "yyyy-MM-dd'T'hh:mm:ss";
```
gidsg: This name and the one after is a bit confusing with the `YEARS` plural. How about instead `YEAR_MONTH_DAY_HOURS_MINUTES_SECONDS`?

src/main/java/uk/gov/register/util/DateFormatChecker.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class DateFormatChecker {
+
+    public static boolean isDateFormatValid(final String dateToValidate) {
+        if (StringUtils.isEmpty(dateToValidate)) {
+            return true;
+        }
+
+        try {
+            DateUtils.parseDate(dateToValidate, YEAR, YEAR_MONTH, YEAR_MONTH_DAY, YEARS_MONTH_DATA_TIME, YEARS_MONTH_DATA_TIME_UTC);
```
gidsg: Should we use [parseDateStrictly](https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/time/DateUtils.html)?

src/main/java/uk/gov/register/service/ItemValidator.java
```diff
@@ -64,6 +66,17 @@ private void validateFieldsValue(JsonNode inputEntry, Map<String, Field> fields)
         });
     }
 
+    private void dataTypeBasedValidation(final JsonNode inputEntry, final String fieldName, final Datatype datatype,
+                                         final JsonNode fieldValue) {
+        if (DataType.DATETIME.getName().equals(datatype.getName())) {
+            if (!DateFormatChecker.isDateFormatValid(fieldValue.textValue())) {
```
karlbaker02: We could pass this conditional check to the existing `throwEntryValidationExceptionIfConditionIsFalse` method, which a) removes a layer of nesting `if` statements here and b) means that we don't an additional `throwEntryValidationException` method

src/main/java/uk/gov/register/util/DateFormat.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.register.util;
+
+public interface DateFormat {
```
karlbaker02: I'd be interested to hear what others think, but to me this feels like we should create these variables as `public static` inside of a class, as opposed to an interface.

src/main/java/uk/gov/register/util/DateFormatChecker.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.register.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.time.DateUtils;
+
+import java.text.ParseException;
+
+import static uk.gov.register.util.DateFormat.ISO_8601.*;
+
+public class DateFormatChecker {
```
karlbaker02: Do we want to test this class also?

src/main/java/uk/gov/register/util/enums/DataType.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.register.util.enums;
+
+public enum DataType {
```
karlbaker02: We have an existing `Datatype` interface which may get confusing with this `enum`. Given that we're storing the pattern separately from this enum, it might be worth doing a comparison of the field's name against `"datetime"` or its type against `DatetimeDatatype` in the ItemValidator instead?

src/test/java/uk/gov/register/service/ItemValidatorTest.java
```diff
@@ -151,4 +167,91 @@ public void validateItem_failsOnEmptyStringField() throws IOException {
             assertThat(e.getEntry().toString(), equalTo(jsonString));
         }
     }
+
+    @Test
+    public void validateItem_date_year() throws IOException {
```
karlbaker02: The general convention for our unit tests is something along the lines of:

`methodName_shouldDoSomething_whenGivenSomeInputDataOrSomeConditionIsTrue`

In the context of this test, for example:

`validateItem_shouldValidateSuccessfully_whenInputDateTimeIsOfFormatYYYYDD`

src/main/java/uk/gov/register/util/enums/DataType.java
gidsg: Yes, I think something like `isinstanceof` `DatatimeDatatype` would make sense

src/main/java/uk/gov/register/util/DateFormat.java
fjavierm: More opinions about this?

src/main/java/uk/gov/register/util/DateFormatChecker.java
fjavierm: I think it's repetitive but I've added them

src/main/java/uk/gov/register/util/DateFormat.java
gidsg: Erm, I think I agree with @karlbaker02 that since there is nothing directly implementing this interface, that it makes more sense as`public static`?

fjavierm: I was just think about a centrilized place to have all the date formats not just a class where to burry them. But, if both of you agree to change it, I'll do it

michaelabenyohai: I don't mind this being in it's own interface but I don't have strong feelings about either solution. Happy with either.

src/main/java/uk/gov/register/service/ItemValidator.java
```diff
@@ -64,6 +66,16 @@ private void validateFieldsValue(JsonNode inputEntry, Map<String, Field> fields)
         });
     }
 
+    private void dataTypeBasedValidation(final JsonNode inputEntry, final String fieldName, final Datatype datatype,
```
michaelabenyohai: I feel that this method should exist in the implementation of the existing `DaterimeDataType.isValid()` method (there is a TODO to implement validation here). That way validation would be consistent for all datatypes as each has an `isValid()` method implemented.

src/main/java/uk/gov/register/util/DateFormat.java
fjavierm: I prefer to have it defined on it's own considering is a format we use in the system

src/main/java/uk/gov/register/service/ItemValidator.java
```diff
@@ -43,25 +44,36 @@ private void validateFields(JsonNode inputEntry, RegisterMetadata registerMetada
...
-            Datatype datatype = field.getDatatype();
-
-            JsonNode fieldValue = inputEntry.get(fieldName);
+            this.validateFieldDataType(fieldName,datatype, inputEntry, fieldValue);
+            this.validateFieldValue(fieldName, field, datatype, inputEntry, fieldValue);
+        });
+    }
 
-            if (field.getCardinality().equals(Cardinality.MANY)) {
+    private void validateFieldDataType(final String fieldName, final Datatype datatype, final JsonNode inputEntry, final JsonNode fieldValue) {
```
michaelabenyohai: Now that you have implemented `DatetimeDatatype.isValid()` I don't think you need this new method called `validateFieldDataType` as `Datatype.isValid()` is an interface method that gets called for all Datatypes (including StringDatatype, IntegerDatatype etc.) in the original existing `validateField` method. I don't think this class requires any changes at all.

src/test/java/uk/gov/register/service/ItemValidatorTest.java
```diff
@@ -151,4 +166,89 @@ public void validateItem_failsOnEmptyStringField() throws IOException {
...
+    @Test
+    public void validateItem_throwsValidationException_whenInputDateTimeIsNotValid() throws IOException {
+        String jsonString = "{\"country\":\"myCountry\",\"start-date\":\"18/12/2009\"}";
+        JsonNode jsonNode = nodeOf(jsonString);
+
+        try {
+            countryItemValidator.validateItem(jsonNode, countryFieldsByName, countryRegisterMetadata);
+            fail("Must not execute this statement");
+        } catch (ItemValidationException e) {
+            assertThat(e.getMessage(), equalTo("Field 'start-date' value must be of type 'datetime'"));
```
fjavierm: I've remove it but, in my opinion, now the error message is misleading. It's complaining that it's not the correct data type, what can be true sometimes (Ex: "value12345') but it is saying the same when the value is a datetime but not in the correct format (Ex: 01/01/2018). What do you think @michaelabenyohai?

michaelabenyohai: I think this is fine because the the error says `value must be datetime` and by `datetime` we mean whatever is defined in https://datatype.register.gov.uk/record/datetime. In this case, it is true that both "value12345" and "01/01/2018" are not a `datetime` as per our register definition.

michaelabenyohai: I do believe that we need to better document these error messages though. This might be something to add to our Tech Docs. I'll speak to our Tech Writer and see.

PR #343, opened 2017-09-14
Implement new design for download/preview

fjavierm:


fjavierm:
Everything apparently working. Now that I have a little bit more knowledge about the cource code, I'm trying to refactor things.

src/main/java/uk/gov/register/resources/PreviewEntriesResource.java
```diff
@@ -0,0 +1,92 @@
...
+@Path("/")
+public class PreviewEntriesResource {
+
+    private static final int START = 0;
+
+    protected final ViewFactory viewFactory;
+    private final RegisterReadOnly register;
+
+    @Inject
+    public PreviewEntriesResource(final ViewFactory viewFactory, final RegisterReadOnly register) {
```
karlbaker02: This is probably more of an individual style point, but do we need to specify `final` here when we assign these variables to `final` class variables?

src/main/java/uk/gov/register/resources/PreviewEntriesResource.java
```diff
@@ -0,0 +1,92 @@
...
+        this.viewFactory = viewFactory;
+        this.register = register;
+    }
+
+    @GET
+    @Path("/preview-entries-json")
+    @Produces(ExtraMediaType.TEXT_HTML)
+    @Timed
+    public View jsonPreview() {
+        return this.previewPageView(MediaType.APPLICATION_JSON);
```
karlbaker02: Again a style point (being consistent with the rest of the codebase), we only use `this` in constructors.

src/main/java/uk/gov/register/resources/PreviewEntriesResource.java
```diff
@@ -0,0 +1,92 @@
...
+    @Path("/preview-entries-ttl/{entry-key}")
+    @Produces(ExtraMediaType.TEXT_HTML)
+    @Timed
+    public View ttlByKeyPreview(@PathParam("entry-key") final int key) {
+        return previewByKeyPageView(key, ExtraMediaType.TEXT_TTL_TYPE.getSubtype());
+    }
+
+    private View previewPageView(final String mediaType) {
+        final int limit = register.getTotalEntries() > IndexSizePagination.ENTRY_LIMIT
+                ? IndexSizePagination.ENTRY_LIMIT + 1
```
karlbaker02: Could you explain why we do the `+1` here?

src/main/java/uk/gov/register/resources/PreviewEntriesResource.java
```diff
@@ -0,0 +1,92 @@
...
+    private final RegisterReadOnly register;
+
+    @Inject
+    public PreviewEntriesResource(final ViewFactory viewFactory, final RegisterReadOnly register) {
+        this.viewFactory = viewFactory;
+        this.register = register;
+    }
+
+    @GET
+    @Path("/preview-entries-json")
```
karlbaker02: I think rather than having three separate endpoints, we have one which is `/preview-entries/{media-type}`

src/main/java/uk/gov/register/resources/PreviewEntriesResource.java
```diff
@@ -0,0 +1,92 @@
...
+    @GET
+    @Path("/preview-entries-ttl")
+    @Produces(ExtraMediaType.TEXT_HTML)
+    @Timed
+    public View ttlPreview() {
+        return previewPageView(ExtraMediaType.TEXT_TTL_TYPE.getSubtype());
+    }
+
+    @GET
+    @Path("/preview-entries-json/{entry-key}")
```
karlbaker02: Again, same here with the three separate endpoints.

src/main/java/uk/gov/register/resources/PreviewItemsResource.java
```diff
@@ -0,0 +1,58 @@
...
+    public PreviewItemsResource(final ViewFactory viewFactory, final RegisterReadOnly register) {
+        this.viewFactory = viewFactory;
+        this.register = register;
+    }
+
+    @GET
+    @Path("/preview-items-json/sha-256:{item-hash}")
+    @Produces(ExtraMediaType.TEXT_HTML)
+    @Timed
+    public View jsonByKeyPreview(@PathParam("item-hash") final String itemHash) {
```
karlbaker02: As in `PreviewEntriesResource`, I'd combine these three separate endpoints into one, taking the `media-type` as a parameter.

src/main/java/uk/gov/register/resources/PreviewRecordsResource.java
```diff
@@ -0,0 +1,92 @@
...
+    private final RegisterReadOnly register;
+
+    @Inject
+    public PreviewRecordsResource(final ViewFactory viewFactory, final RegisterReadOnly register) {
+        this.viewFactory = viewFactory;
+        this.register = register;
+    }
+
+    @GET
+    @Path("/preview-records-json")
```
karlbaker02: And again

src/main/java/uk/gov/register/resources/PreviewRecordsResource.java
```diff
@@ -0,0 +1,92 @@
...
+        final int limit = register.getTotalRecords() > IndexSizePagination.ENTRY_LIMIT
+                ? IndexSizePagination.ENTRY_LIMIT
+                : register.getTotalRecords();
+        final List<Record> records = register.getRecords(limit, DEFAULT_OFFSET);
+
+        return viewFactory.previewRecordsPageView(records, null, mediaType);
+    }
+
+    private View previewByKeyPageView(final String key, final String mediaType) {
+        final List<Record> records = Collections.singletonList(register.getRecord(key).orElseThrow(NotFoundException::new));
```
karlbaker02: We should be consistent in the errors that we're displaying. In `PreviewItemResource` we display more information to the user, telling them that an item with the key they searched for could not be found, whereas here we give no clue to the user as to what went wrong.

src/main/java/uk/gov/register/thymeleaf/ThymeleafView.java
```diff
@@ -21,17 +17,21 @@
 public class ThymeleafView extends View {
     protected final RequestContext requestContext;
     private final RegisterResolver registerResolver;
-    private final RegisterReadOnly register;
-    private Optional<String> registerTrackingId;
+    protected final RegisterReadOnly register;
```
karlbaker02: This isn't being used by any subclass of `ThymeleafView` - does it need to be `protected`?

src/main/java/uk/gov/register/thymeleaf/ThymeleafView.java
```diff
@@ -21,17 +17,21 @@
...
 
-    public ThymeleafView(RequestContext requestContext, String templateName, RegisterTrackingConfiguration registerTrackingConfiguration, RegisterResolver registerResolver, RegisterReadOnly register) {
+    public ThymeleafView(final RequestContext requestContext, final String templateName, final RegisterTrackingConfiguration registerTrackingConfiguration, final RegisterResolver registerResolver, final RegisterReadOnly register) {
         super(templateName, StandardCharsets.UTF_8);
         this.requestContext = requestContext;
-        this.registerTrackingId = registerTrackingConfiguration.getRegisterTrackingId();
+        registerTrackingId = registerTrackingConfiguration.getRegisterTrackingId();
         this.registerResolver = registerResolver;
         this.register = register;
+        totalRecords = register.getTotalRecords();
```
karlbaker02: `totalRecords` and `totalEntries` are very view-specific, in that they are only being used by the preview pages and the homepage. I don't think we should be including them as properties on this base view.

src/main/java/uk/gov/register/util/ElementType.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.register.util;
+
+public enum ElementType {
```
karlbaker02: I can see why you have this, but I wonder if we should stick with our current pattern of having one page for multiple records/entries/items and another page for a single record/entry/item. This is probably worth some discussion.

src/main/java/uk/gov/register/views/EntryListView.java
```diff
@@ -1,36 +1,109 @@
...
+    public String entriesTo(final String mediaType, final Provider<RegisterName> registerNameProvider) {
+        final ByteArrayOutputStream outputStream;
+        final RegisterResolver registerResolver;
+        final EntryListTurtleWriter entryTurtleWriter;
+        String registerInTextFormatted = StringUtils.EMPTY;
+
+        try {
+            if (ExtraMediaType.TEXT_TTL_TYPE.getSubtype().equals(mediaType)) {
+                outputStream = new ByteArrayOutputStream();
+                registerResolver = registerName -> URI.create("http://" + registerName + ".register.gov.uk");
```
karlbaker02: This will only work for beta registers.

src/main/java/uk/gov/register/views/EntryListView.java
```diff
@@ -1,36 +1,109 @@
...
+            } else if (ExtraMediaType.TEXT_YAML_TYPE.getSubtype().equals(mediaType)) {
+                registerInTextFormatted = yamlObjectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(getNestedEntryJson());
+            } else {
+                registerInTextFormatted = jsonObjectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(getNestedEntryJson());
+            }
+        } catch (final IOException ex) {
+            LOGGER.error("Error processing preview request. Impossible process the register items");
+        }
+
+        return StringEscapeUtils.escapeHtml(registerInTextFormatted.isEmpty() ? registerInTextFormatted : END_OF_LINE + registerInTextFormatted);
```
karlbaker02: Two questions: why `\n` if `registerInTextFormatted` is not empty (as opposed to the view taking care of this) and do we notify the user there was an error if there was an error?

src/main/java/uk/gov/register/views/ItemView.java
```diff
@@ -28,16 +49,45 @@ public ItemView(HashValue sha256hex, Map<String, FieldValue> fieldValueMap, Iter
...
+            } else if (ExtraMediaType.TEXT_YAML_TYPE.getSubtype().equals(mediaType)) {
+                registerInTextFormatted = yamlObjectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(getContent());
+            } else {
+                registerInTextFormatted = jsonObjectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(getContent());
+            }
+        } catch (final IOException ex) {
+            LOGGER.error("Error processing preview request. Impossible process the register items");
+        }
+
+        return StringEscapeUtils.escapeHtml(registerInTextFormatted.isEmpty() ? registerInTextFormatted : END_OF_LINE + registerInTextFormatted);
```
karlbaker02: Same two questions: why \n if registerInTextFormatted is not empty (as opposed to the view taking care of this) and do we notify the user there was an error if there was an error?

src/main/java/uk/gov/register/views/PreviewPageView.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return String.format(elementType.getMultipleElementsDownloadLocation(), getPreviewTypeTag().toLowerCase());
+    }
+
+    @SuppressWarnings("unused, used by templates")
+    public String getUrlSingleElements() {
+        return String.format(elementType.getSingleElementDownloadLocation(), getKey(), getPreviewTypeTag().toLowerCase());
+    }
+
+    public abstract int getTotalObjects();
```
karlbaker02: What is total objects?

src/main/java/uk/gov/register/views/PreviewPageView.java
```diff
@@ -0,0 +1,117 @@
...
+                           final ElementType elementType,
+                           final HomepageContent homepageContent,
+                           final Provider<RegisterName> registerNameProvider) {
+        super(requestContext, PREVIEW_HTML_PAGE, registerTrackingConfiguration, registerResolver, register);
+        this.key = key;
+        this.previewType = previewType;
+        this.elementType = elementType;
+        this.homepageContent = homepageContent;
+        this.registerNameProvider = registerNameProvider;
+        totalRecords = register.getTotalRecords();
```
karlbaker02: Could we just call these methods down below in the `getTotalRecords` and `getTotalEntries` methods rather than creating new properties for them?

src/main/java/uk/gov/register/views/ItemView.java
fjavierm: Same two answers.
1. Because the content of the <pre> element in the view needs the \n to show properly the first line due to the JS modifications prismjs does.
2. No, we just return empty. I haven't seen any mechanism of returning proper errors in the views. If there is one, can you point me to it?

src/main/java/uk/gov/register/views/PreviewPageView.java
fjavierm: Actaully, no. At least without modifiying the class `ThymeleafView` and exposing the variable `registers` what at this point looks more invasive than just store the values in the class it needs them. In addtion, there is a naming problem if this is done because the `getRegister` is not returning the `register` object, it's returning the `RegisterMetadata` object.

src/main/java/uk/gov/register/views/PreviewPageView.java
fjavierm: A generic method to recover the number of objects available in the view, for records return number of records, for entries, number of entries and for the item 1

src/main/java/uk/gov/register/views/ViewFactory.java
```diff
@@ -27,17 +29,19 @@
...
+    public ViewFactory(final RequestContext requestContext,
+                       final PublicBodiesConfiguration publicBodiesConfiguration,
+                       final GovukOrganisationClient organisationClient,
+                       final RegisterDomainConfiguration registerDomainConfiguration,
+                       final Provider<HomepageContentConfiguration> homepageContentConfiguration,
+                       final Provider<RegisterTrackingConfiguration> registerTrackingConfiguration,
+                       final RegisterResolver registerResolver,
+                       final Provider<RegisterReadOnly> register,
+                       final Provider<RegisterLinkService> registerLinkService, final ItemConverter itemConverter,
+                       final Provider<RegisterName> registerNameProvider) {
```
karlbaker02: `RegisterName` is available from `RegisterReadOnly` - is this because you need it inside of a `Provider`?

src/main/java/uk/gov/register/views/ViewFactory.java
```diff
@@ -87,66 +92,106 @@ public HomePageView homePageView(int totalRecords, int totalEntries, Optional<In
...
+    public RecordsView getRecordsMediaView(final List<Record> records) {
         return new RecordsView(records, register.get().getFieldsByName(), itemConverter, false, false);
     }
 
-    public RecordsView getIndexRecordsMediaView(List<Record> records) {
+    public RecordsView getIndexRecordsMediaView(final List<Record> records) {
         return new RecordsView(records, register.get().getFieldsByName(), itemConverter, false, true);
     }
 
+    public PreviewRecordPageView previewRecordsPageView(final List<Record> records, final String key, final String previewType) {
```
karlbaker02: Let's follow the convention here and have a `get` prefix

build.gradle
```diff
@@ -27,8 +27,8 @@ def assetsDir = new File(projectDir, "src/main/resources/assets")
 
 compass {
     importPath = files(
-        new File(assetsDir, "govuk_frontend_toolkit/stylesheets"),
-        new File(assetsDir, "govuk_template/source/assets/stylesheets")
+            new File(assetsDir, "govuk_frontend_toolkit/stylesheets"),
```
danmitchell-: Is this indentation right? 2 spaces?

src/main/resources/assets/sass/main.scss
```diff
@@ -300,4 +301,12 @@ dl {
     padding-bottom: 1em;
     width: 65%;
   }
-}
\ No newline at end of file
+}
+
+.preview-box {
+    border: 1px solid #d7d7d7;
```
danmitchell-: We should use a colour variables here. i.e. `$grey-2`

src/main/resources/templates/entries.html
```diff
@@ -3,76 +3,76 @@
 <html xmlns:th="http://www.thymeleaf.org" xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
 <head th:include="fragments/head.html::head"></head>
 <body th:class="${isGovukBranded ? 'gov-uk' : 'openregister-org'}">
-  <header th:replace="fragments/global-header.html::global-header"></header>
+<header th:replace="fragments/global-header.html::global-header"></header>
```
danmitchell-: I think the indentation is off in this file too.

src/main/java/uk/gov/register/views/ViewFactory.java
fjavierm: The provider is needed in the different TurtleWriter constructors

src/main/java/uk/gov/register/views/ViewFactory.java
fjavierm: You can find both cases in the file. Ex: `downloadPageView`

build.gradle
fjavierm: We don't have two spaces in gradle, we have 1 tab (4 spaces) but in this case is something embedeed, for this reasons is 2 tabs.

src/main/java/uk/gov/register/resources/PreviewEntriesResource.java
fjavierm: This was already discussed. Final stay

karlbaker02: Discussed.

src/main/java/uk/gov/register/util/ElementType.java
karlbaker02: Discussed.

src/main/java/uk/gov/register/views/EntryListView.java
karlbaker02: Discussed.

src/main/java/uk/gov/register/views/ItemView.java
karlbaker02: Discussed.

src/main/java/uk/gov/register/views/PreviewPageView.java
karlbaker02: Discussed.

src/main/java/uk/gov/register/views/PreviewPageView.java
karlbaker02: Discussed.

src/main/java/uk/gov/register/views/ViewFactory.java
karlbaker02: Discussed.

PR #344, opened 2017-09-18
Records page times out when getting 5000 records.

fjavierm:
Mainly the structure used to populate the data en in the view was to complex to loop through it. In addition, it was consuming a lot of memory. The data estructure used in the view has been simplified.


gidsg:
Would be good to explain in the description why it was previously timing out.

PR #345, opened 2017-09-18
Increase instance size to 4GB

michaelabenyohai:
Before moving to PaaS the size of each AWS EC2 instance was 4GB.
This was reduced to 1GB when we moved to PaaS but for no real
reason. Some instances are almost at their memory limit and
it's suspected that this could be causing some issues with
loading and querying data.

### DIFF
--- a/deploy/manifests/alpha/basic.yml
+++ b/deploy/manifests/alpha/basic.yml
@@ -1,7 +1,7 @@
 ---
 applications:
 - name: alpha-basic
-  memory: 1024M
+  memory: 4GB
   instances: 2
   buildpack: java_buildpack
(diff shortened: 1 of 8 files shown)


karlbaker02:
If we wanted to quickly test we could override the memory allocation in the `manifest` by specifying the `-m` option on the command line (https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#memory). We should restore our memory requirements to what they were previously in any case.

michaelabenyohai:
@karlbaker02 good tip! I will merge anyway. I suspect we need to do some analysis as to what these should be but I think they are dangerously low currently.

Issue #342, opened 2017-09-14
Running application from scratch fails


Glamdring:
Hello,

I tried running the openregister-java application locally, using `sudo ENVIRONMENT=beta REGISTERS=country ./run-application.sh` but it failed to start.

Initially it reports:

```
Waiting for HTTP on :8081
Loading register...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4491    0  4491    0     0  12452      0 --:--:-- --:--:-- --:--:-- 12475
curl: (22) The requested URL returned error: 400 Bad Request
Stopping and removing containers...
```
Then I added --verbose to the curl command that fetches the fsf in run-application.sh:

`--data-binary @<(curl --verbose "https://$register.$DOMAIN/download-rsf")`

And that reported:

```
* Connection #0 to host field.register.gov.uk left intact
{"success":false,"message":"Exception when executing command: RegisterCommand{commandName='append-entry', arguments=[system, register:field, 2016-08-04T14:45:41Z, sha-256:ebb8bc0c3639577c5b7d01e5b5a382c52dfc7dee41c290bd8b29647db00259d4]}","details":"Definition of register field does not match Register Register"}S
```

How can this be addressed and fixed?

As a sidenote: maybe explicitly point out in the readme that when cloning the repo, one should also init the submodules, otherwise the gradle build fails (as it doesn't find the govuk templates)

fjavierm:
Hello,

The issue has been fixed. You can take a look to the PR #346.

Basically, the proper values for the beta environment have been set. And some intructions have been added to the readme to run it in environments other than beta.

gidsg:
Hi @Glamdring 
Thanks for reporting this issue. We pushed a fix for this in #346 please try again after pulling master.

PR #346, opened 2017-09-19
Running application from scratch fails

fjavierm:
Fixes issue #342.
Assign proper default values for beta environments and update the README documentation about how to run it locally.

### DIFF
--- a/RUNNING_LOCALLY.md
+++ b/RUNNING_LOCALLY.md
@@ -50,6 +50,26 @@ Change `register: country` to `register: school-type`.
 
 Update the schema in the same way, to `schema: school-type`.
 
+Update fields and registers location for environments other than beta in the same way.
+
+Fields locations per environment:
+
+Variable: `fieldsYamlLocation`
+
+Possible values:
+* alpha: https://field.alpha.openregister.org/records.yaml
+* beta: https://field.register.gov.uk/records.yaml
+* discovery: https://field.discovery.openregister.org/records.yaml
+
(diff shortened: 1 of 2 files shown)


gidsg:
You should put in the description fixes #342. This will keep it in sync with the corresponding github issue see: https://github.com/blog/1506-closing-issues-via-pull-requests

RUNNING_LOCALLY.md
```diff
@@ -50,6 +50,24 @@ Change `register: country` to `register: school-type`.
...
+Possible values:
+alpha: https://field.alpha.openregister.org/records.yaml
+beta: https://field.register.gov.uk/records.yaml
+discovery: https://field.discovery.openregister.org/records.yaml
+
+Registers locations per environment:
+
+Variable: `registersYamlLocation`
+Possible values:
+alpha: https://register.alpha.openregister.org/records.yaml
```
gidsg: Can you prefix these with `*` to make them bullet points? Right now they run in one line when rendered.

RUNNING_LOCALLY.md
```diff
@@ -50,6 +50,24 @@ Change `register: country` to `register: school-type`.
 
 Update the schema in the same way, to `schema: school-type`.
 
+Update fields and registers location for environments different to beta in the same way.
```
gidsg: it is a very minor wording thing, but I would rather say _'other than'_ instead of _'different to'_

PR #350, opened 2017-09-22
implement 'spreadsheets' as a spreadsheet format

fjavierm:
Adding the possibility to download the records/entries/items in XLSX format.

### DIFF
--- a/src/main/java/uk/gov/register/views/representations/spreadsheet/EntriesSpreadSheetWriter.java
+++ b/src/main/java/uk/gov/register/views/representations/spreadsheet/EntriesSpreadSheetWriter.java
@@ -1,0 +1,67 @@
+package uk.gov.register.views.representations.spreadsheet;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.poi.ss.usermodel.Workbook;
+import uk.gov.register.util.HashValue;
+import uk.gov.register.views.EntryListView;
+import uk.gov.register.views.representations.ExtraMediaType;
+import uk.gov.register.views.representations.RepresentationWriter;
+
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Provider
+@Produces(ExtraMediaType.APPLICATION_SPREADSHEET)
+public class EntriesSpreadSheetWriter extends RepresentationWriter<EntryListView> {
+
+    @Override
+    public void writeTo(final EntryListView entryListView, final Class<?> type, final Type genericType,
+                        final Annotation[] annotations, final MediaType mediaType,
+                        final MultivaluedMap<String, Object> httpHeaders, final OutputStream entityStream)
+            throws IOException, WebApplicationException {
+        final List<String> fieldNames = new ArrayList<>();
+        final List<Map<String, String>> elements = new ArrayList<>();
+        final Workbook workbook;
+
+        fieldNames.add("index-entry-number");
+        fieldNames.add("entry-number");
+        fieldNames.add("entry-timestamp");
+        fieldNames.add("key");
+        fieldNames.add("item-hash");
+
+        entryListView.getEntries().forEach(entry -> {
+            final Map<String, String> element = new HashMap<>();
+            String itemHashValues = StringUtils.EMPTY;
+
+            element.put("index-entry-number", entry.getIndexEntryNumber().toString());
+            element.put("entry-number", entry.getEntryNumber().toString());
+            element.put("entry-timestamp", entry.getTimestampAsISOFormat());
+            element.put("key", entry.getKey());
+
+            if (!entry.getItemHashes().isEmpty()) {
+                itemHashValues = entry.getItemHashes()
+                        .stream().map(HashValue::toString).collect(Collectors.joining(";"));
+            }
+
+            element.put("item-hash", itemHashValues);
+
+            elements.add(element);
+        });
+
+        workbook = WorkbookGenerator.toSpreadSheet("entries", fieldNames, elements);
+        workbook.write(entityStream);
+        entityStream.flush();
+    }
+}
--- a/src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadSheetWriter.java
+++ b/src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadSheetWriter.java
@@ -1,0 +1,27 @@
+package uk.gov.register.views.representations.spreadsheet;
+
+import uk.gov.register.views.RecordView;
+import uk.gov.register.views.representations.ExtraMediaType;
+import uk.gov.register.views.representations.RepresentationWriter;
+
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@Produces(ExtraMediaType.APPLICATION_SPREADSHEET)
+public class RecordSpreadSheetWriter extends RepresentationWriter<RecordView> implements RecordSpreadsheet {
+
+    @Override
+    public void writeTo(final RecordView recordView, final Class<?> type, final Type genericType, final Annotation[] annotations,
+                        final MediaType mediaType, final MultivaluedMap<String, Object> httpHeaders,
+                        final OutputStream entityStream) throws IOException, WebApplicationException {
+        process(recordView.getFields(), recordView.getRecords(), "record", entityStream);
+    }
+}
--- a/src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadsheet.java
+++ b/src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadsheet.java
@@ -1,0 +1,49 @@
+package uk.gov.register.views.representations.spreadsheet;
+
+import org.apache.poi.ss.usermodel.Workbook;
+import uk.gov.register.core.Entry;
+import uk.gov.register.core.Field;
+import uk.gov.register.views.ItemView;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public interface RecordSpreadsheet {
+
+    default void process(final Iterable<Field> fields, final Map<Entry, List<ItemView>> records, final String sheetLabel,
+                         final OutputStream entityStream) throws IOException {
+        final List<String> fieldNames = new ArrayList<>();
+        final List<Map<String, String>> elements = new ArrayList<>();
+        final Workbook workbook;
+
+        fieldNames.add("index-entry-number");
+        fieldNames.add("entry-number");
+        fieldNames.add("entry-timestamp");
+        fieldNames.add("key");
+
+        fields.forEach(field -> fieldNames.add(field.fieldName));
+
+        records.forEach((key, items) ->
+            items.forEach(item -> {
+                final Map<String, String> element = new HashMap<>();
+
+                element.put("index-entry-number", key.getIndexEntryNumber().toString());
+                element.put("entry-number", key.getEntryNumber().toString());
+                element.put("entry-timestamp", key.getTimestampAsISOFormat());
+                element.put("key", key.getKey());
+
+                item.getContent().forEach((contentKey, contentValue) -> element.put(contentKey, contentValue.getValue()));
+
+                elements.add(element);
+            })
+        );
+
+        workbook = WorkbookGenerator.toSpreadSheet(sheetLabel, fieldNames, elements);
+        workbook.write(entityStream);
+        entityStream.flush();
+    }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,7 @@ configurations {
 }
 
 dependencies {
-    compile dropwizard, stringTemplate, apacheCommonsCodec, apacheCommonsCollections, apacheCommonsLang3, awsS3
+    compile dropwizard, stringTemplate, apacheCommonsCodec, apacheCommonsCollections, apacheCommonsLang3, apachePoi, apachePoiOoxml, awsS3
     compile jacksonCsv, jena, jerseyMedia, markdown, thymeleaf, verifiableLog
     compile jdbi, jersey_client
     compile dropwizardLogstash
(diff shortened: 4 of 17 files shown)


src/main/java/uk/gov/register/service/SpreadSheetTransformerService.java
```diff
@@ -0,0 +1,97 @@
...
+        itemView.getFields().forEach(field -> fieldNames.add(field.getFieldName()));
+
+        itemView.getContent().forEach((fieldName, fieldValue) -> element.put(fieldName, fieldValue.getValue()));
+
+        elements.add(element);
+
+        toSpreadSheet(outputStream, registerName, fieldNames, elements);
+    }
+
+    public String getFilename(final String identifier) {
```
karlbaker02: If we made the endpoint `/*. xlsx` would we need to do this? If we follow our other downloads at the moment, none of them have any information regarding register / timestamp.

fjavierm: We want to added, right now, if you download multiple files, yo don't have any clue about what they are if you don't open them.

src/main/java/uk/gov/register/core/Field.java
```diff
@@ -47,16 +51,28 @@ public String getText() {
     }
 
     @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+    public boolean equals(final Object o) {
+        if (this == o) {
```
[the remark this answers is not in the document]
fjavierm: We don't do one-liners in any other places (getters, setters are common suspects), why do it here?

In fact, now that we are using the common-lang3 we can use the HashCodeBuilder and the EqualsBuilder that help a lot with readability.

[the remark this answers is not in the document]
karlbaker02: It's true. I think the reason we do one-liners here (and I may be wrong about this) is because we have a lot of individual checks which are small enough that they can be one-liners, and that by doing this we keep the code compact (albeit slightly harder to read).

I'm generally in favour of not having one-liners, but you have to read through a lot more of the code to get to the end in this example. The one-liners also give me the impression at least that the method is doing some fairly simple checks, as they aren't complex enough to need to span multiple lines, and makes it easier to get a sense of the flow of success or failure of the method.

src/main/java/uk/gov/register/providers/DownloadExceptionMapper.java
```diff
@@ -0,0 +1,15 @@
...
+import uk.gov.register.exceptions.DownloadException;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class DownloadExceptionMapper implements ExceptionMapper<DownloadException> {
+    @Override
+    public Response toResponse(final DownloadException exception) {
```
karlbaker02: I might steal one of Michaela's comments from another PR here and ask *can we use the ExceptionView to display this error?*

src/main/java/uk/gov/register/views/representations/YamlWriter.java
```diff
@@ -26,7 +26,7 @@ public YamlWriter() {
     }
 
     @Override
-    public void writeTo(Object object, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream) throws IOException, WebApplicationException {
+    public void writeTo(final Object object, final Class<?> type, final Type genericType, final Annotation[] annotations, final MediaType mediaType, final MultivaluedMap<String, Object> httpHeaders, final OutputStream entityStream) throws IOException, WebApplicationException {
```
karlbaker02: Just as a general note (and I know this is from the auto-formatter), I don't think we should be including files where the actual file hasn't changed in some meaningful way related to the problem the PR is trying to address.

src/main/java/uk/gov/register/views/representations/spreadsheet/EntriesSpreadSheetWriter.java
```diff
@@ -0,0 +1,67 @@
...
+            final Map<String, String> element = new HashMap<>();
+            String itemHashValues = StringUtils.EMPTY;
+
+            element.put("index-entry-number", entry.getIndexEntryNumber().toString());
+            element.put("entry-number", entry.getEntryNumber().toString());
+            element.put("entry-timestamp", entry.getTimestampAsISOFormat());
+            element.put("key", entry.getKey());
+
+            if (!entry.getItemHashes().isEmpty()) {
+                itemHashValues = entry.getItemHashes()
```
karlbaker02: @fjavierm I may have told you the wrong thing here. @michaelabenyohai what do we want to do in a spreadsheet if there are either multiple or no item hashes for an entry? The specification does not specify for this format, nor for csv or tsv.

src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadsheet.java
```diff
@@ -0,0 +1,48 @@
...
+
+        records.forEach((key, items) -> {
+            final Map<String, String> element = new HashMap<>();
+
+            element.put("index-entry-number", key.getIndexEntryNumber().toString());
+            element.put("entry-number", key.getEntryNumber().toString());
+            element.put("entry-timestamp", key.getTimestampAsISOFormat());
+            element.put("key", key.getKey());
+
+            items.forEach(item -> item.getContent()
```
karlbaker02: A potential problem here (probably unlikely to occur) would be if we had a property in the item with the same name, i.e. `key`. We could avoid this by namespacing the item content, but this might be a step too far for our needs at the moment.

src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadSheetWriter.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@Produces(ExtraMediaType.TEXT_SPREADSHEET)
+public class RecordSpreadSheetWriter extends RepresentationWriter<RecordView> implements RecordSpreadsheet {
```
karlbaker02: This class is identical to `RecordsSpreadSheetWriter` in all respects except the different in the view that the classes take as a parameter (but the fact that we only use this view to call `getRecords()` means there is no real difference). Did you mean to pass a parameter of `"records"` to the `process` method here, as opposed to `"record"`? If not, do we need both classes?

karlbaker02:
I think that we might need to reinstate, somewhere, the `Content-Disposition` for `xlsx` to indicate that it is an attachment and not to be displayed inline?

karlbaker02:
Do we want to support the `/record/{record-key}/entries` endpoint located in `RecordResource` too? It supports all of the other existing data formats

src/main/java/uk/gov/register/core/Field.java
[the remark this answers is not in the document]
fjavierm: I just think we should use the same format for everything not just depending of it is simple all not, specially, because this 'simple' classification can be very subjective. But, we can discuss this us a team. Anyway, I'm going to disconnect the auto-format and save actions.

src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadSheetWriter.java
fjavierm: We need both classes because one receives a `RecordsView` and the other one `RecordView`. True that both f them does the same but the signature of the methods `writeTo` are different, and this is the "black magic" done by Dropwizard and the writers. It will be great to be able to pass the parameter `records` and have the behaviour in `RecordSpreadsheet` in one of this classes and delete the other one, but apparently it doesn't work that way.
And, if you observe the turtle writer there are different classes for Records and Record too.

src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadsheet.java
fjavierm: I agree, field names must be unique but, as you have said, for now, application fields have very specific names and I hope there is no going to be conflicts. If we find conflicts  a different approach can be taken.

src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadSheetWriter.java
karlbaker02: Got it. I thought of this point as I saw that `RecordView` is a subclass of `RecordsView`, so wasn't sure by that logic if we would have been able to get away with only one writer.

fjavierm:
About `Content-Disposition`. Not possible to assign now this header because now the writer is in charge of creating a naming the file.

I have tried to set the header `application/force-download` but when the Chrome plugin open the CSV and spreadsheet files now makes a mess with long numeric values that doesn't evaluate as string but as a long number showing scientific notation for them.

fjavierm:
@karlbaker02 Are you sure that the download problem is not because of the plugin?

karlbaker02:
@fjavierm I have tested on Safari and encounter the same issue.

fjavierm:
@karlbaker02 It should be solved now. Try it again please

karlbaker02:
Tested and verified that download for `xlsx` is now working correctly on Chrome and Safari. Only thing to do now is to address the comment for `EntriesSpreadSheetWriter.java` - I'll aim to get an answer on this today.

src/main/java/uk/gov/register/views/representations/spreadsheet/EntriesSpreadSheetWriter.java
karlbaker02: After checking the existing code and the output of CSV files with multiple items per entry, I believe this code is correct. See the attached screenshots.

*Existing code rendering `csv`*

screen shot 2017-09-28 at 14 28 22

*New code rendering `xlsx`*

screen shot 2017-09-28 at 14 28 04

src/main/java/uk/gov/register/views/representations/spreadsheet/RecordSpreadsheet.java
karlbaker02: :+1:

karlbaker02:
Picking up the point from @michaelabenyohai about records, we're currently not maintaining what we do in `csv` in `xlsx` when an entry has multiple items.

In the following screenshots, we're loading a register with 4 entries for 5 distinct items.

Loaded RSF:
```
assert-root-hash	sha-256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
add-item	{"name":"local-authority-eng"}
add-item	{"cardinality":"1","datatype":"string","field":"local-authority-eng","phase":"beta","register":"local-authority-eng","text":"The local authority's ISO 3166-1 alpha3 code. Unique codes have been created for local authorities that don't have an existing ISO code."}
add-item	{"cardinality":"1","datatype":"string","field":"name","phase":"beta","text":"The commonly-used name of a record."}
add-item	{"cardinality":"1","datatype":"string","field":"official-name","phase":"beta","text":"The official or technical name of a record."}
add-item	{"custodian":"Stephen McAllister"}
add-item	{"fields":["local-authority-eng","local-authority-type","name","official-name","start-date","end-date"],"phase":"beta","register":"local-authority-eng","registry":"department-for-communities-and-local-government","text":"Local authorities in England"}
add-item	{"cardinality":"1","datatype":"string","field":"local-authority-type","phase":"beta","register":"local-authority-type","text":"The type of local government organisation."}
add-item	{"cardinality":"1","datatype":"datetime","field":"end-date","phase":"beta","text":"The date a record stopped being applicable. For example, the date a school closed down."}
add-item	{"cardinality":"1","datatype":"datetime","field":"start-date","phase":"beta","text":"The date a record first became relevant to a register. For example, the date a country was first recognised by the UK."}
add-item	{"local-authority-eng":"BBD","local-authority-type":"UA","name":"Blackburn with Darwen","official-name":"Blackburn with Darwen Borough Council","start-date":"1998-04-01"}
add-item	{"local-authority-eng":"BIR","local-authority-type":"MD","name":"Birmingham","official-name":"Birmingham City Council","start-date":"1996-07-01"}
add-item	{"local-authority-eng":"BDF","local-authority-type":"UA","name":"Bedford","official-name":"Bedford Borough Council","start-date":"2009-04-01"}
append-entry	system	name	2017-07-17T10:59:48Z	sha-256:41ebaf26a5a1ae11d2d6f8dce81922ae46ab354b049efdb5d2fc8e2fa951097d
append-entry	system	custodian	2017-07-17T10:59:48Z	sha-256:1a79573f28a473bc281ae5ef7f1a910da710648bc666abf5930e05e1ed962f39
append-entry	system	field:local-authority-eng	2017-01-10T17:16:07Z	sha-256:9980619a9bb4f9badd0456d7f3740fb6d97d7ca816e4db8d34d544e2695b51fe
append-entry	system	field:local-authority-type	2017-01-10T17:16:07Z	sha-256:c244c74c48c14b08aa98346ee8064743e65d6331b09d524c03a333468f3b58da
append-entry	system	field:name	2017-01-10T17:16:07Z	sha-256:a7a9f2237dadcb3980f6ff8220279a3450778e9c78b6f0f12febc974d49a4a9f
append-entry	system	field:official-name	2017-01-10T17:16:07Z	sha-256:5c4728f439f6cbc6c7eea42992b858afc78c182962ba35d169f49db2c88e1e41
append-entry	system	field:start-date	2017-01-10T17:16:07Z	sha-256:1cff4c622577fabd35917760b04bc304af3b950e19735b60152c19f18b55e75e
append-entry	system	field:end-date	2017-01-10T17:16:07Z	sha-256:a557fa9bf4a6ce1991717b9429406f4fa724fb8818dfe018d306694de22efbf7
append-entry	system	register:local-authority-eng	2016-10-21T15:20:42Z	sha-256:c1762331b777d5f820735a693cc9443eebefc076fa7f2ce0ceaeea9d0f52a5d4
append-entry	user	A	2016-10-21T16:11:20Z	sha-256:37dd060a3efa6d5bf16f876e08fa867210f424e2e4a7989d0322218f6772332d
append-entry	user	B	2016-10-21T16:11:20Z	sha-256:01cda95d102807943d68f71bbe7b9b290dec2ebdb15967ca66820f825ece5831
append-entry	user	C	2016-10-21T16:11:20Z	sha-256:bdc17a29807f47a94cf612abf92fcadd2d83176f9ea419b4eeda23af2cf25ef9
add-item	{"local-authority-eng":"LUT","local-authority-type":"UA","name":"Luton","official-name":"Luton Borough Council","start-date":"1905-06-19"}
add-item	{"local-authority-eng":"WOT","local-authority-type":"NMD","name":"Worthing","official-name":"Worthing Borough Council","start-date":"1905-06-25"}
append-entry	user	E	2016-10-21T16:11:20Z	sha-256:304949d78e0c42591cf9bb547bc7143b869bb29217618633e3a9ba933eb1e1d7;sha-256:f3ca6a2079d74458ea5580ce3a214a46fed27a4fdfda8249083ef33190a9d3ef
```

Outputted `csv`:

screen shot 2017-09-28 at 15 46 35

Outputted `xlsx`:

screen shot 2017-09-28 at 15 48 51

We should ensure that `xlsx` matches `csv`.

fjavierm:
@karlbaker02 Fixed, now each item is going to have a row. Quick question. Can a record has no item? As far as I understand, this situation cannot happen but, maybe, I'm wrong

karlbaker02:
Confirmed that records is fixed.

PR #356, opened 2017-09-26
Check that exceptions from ItemConverter do not affect all registers

fjavierm:
Catching exception when there is an error converting the data.

Replacing message `500 Error` with `[ERROR] Not possible to process the field (name, data-type, value): organisation, curie, government-organisationD10.`

### DIFF
--- a/src/main/java/uk/gov/register/exceptions/FieldConversionException.java
+++ b/src/main/java/uk/gov/register/exceptions/FieldConversionException.java
@@ -1,0 +1,7 @@
+package uk.gov.register.exceptions;
+
+public class FieldConversionException extends RuntimeException {
+    public FieldConversionException(final String message) {
+        super(message);
+    }
+}
--- a/src/main/java/uk/gov/register/providers/ThrowableExceptionMapper.java
+++ b/src/main/java/uk/gov/register/providers/ThrowableExceptionMapper.java
@@ -3,11 +3,11 @@ package uk.gov.register.providers;
 import org.glassfish.jersey.server.ParamException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.register.exceptions.FieldConversionException;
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/register/exceptions/FieldConversionException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.register.exceptions;
+
+public class FieldConversionException extends RuntimeException {
```
michaelabenyohai: Is it worth renaming this to be `ItemConversionException` and `ItemConversionExceptionMapper`? It might be more consistent with the name `ItemConverter`, but I'm also happy for this to be left as it is.

src/main/java/uk/gov/register/providers/FieldConversionExceptionMapper.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class FieldConversionExceptionMapper implements ExceptionMapper<FieldConversionException> {
+
+    @Override
+    public Response toResponse(final FieldConversionException exception) {
+        return Response.status(500).entity("[ERROR] " + exception.getMessage()).build();
```
michaelabenyohai: Can we use the `ExceptionView` to display this error? Look at the `ThrowableExceptionMapper` as an example, where it uses `.entity(viewFactory.exceptionServerErrorView())`. We should also do this for "register undefined" but we will do that separately.

src/main/java/uk/gov/register/exceptions/FieldConversionException.java
michaelabenyohai: As discussed, happy for this to keep the current name.

PR #357, opened 2017-09-26
'turtle' URLs to 'ttl' URLs

fjavierm:
URLs in the HTML pages are using the label 'turtle', they should be using 'ttl' that is the supported format and in this way be more intuitive.


gidsg:
Please add a description to describe why we are doing this!

karlbaker02:
@fjavierm @gidsg I don't see where this description is. Can it be added to the commit please?

fjavierm:
It was in the card but, I've added here too

karlbaker02:
Cool. Let's make sure to add the descriptions to the actual commits in future too, as otherwise you have to track down the related PR and makes the history harder to find.

PR #358, opened 2017-09-27
Implement CURIE checking on load

fjavierm:


src/main/java/uk/gov/register/service/ItemValidator.java
```diff
@@ -16,58 +16,59 @@
...
-
                 throwEntryValidationExceptionIfConditionIsFalse(!fieldValue.isArray(), inputEntry, String.format("Field '%s' has cardinality 'n' so the value must be an array of '%s'", fieldName, datatype.getName()));
 
                 fieldValue.elements().forEachRemaining(element -> throwEntryValidationExceptionIfConditionIsFalse(!datatype.isValid(element), inputEntry, String.format("Field '%s' values must be of type '%s'", fieldName, datatype.getName())));
-
             } else {
                 throwEntryValidationExceptionIfConditionIsFalse(!datatype.isValid(fieldValue), inputEntry, String.format("Field '%s' value must be of type '%s'", fieldName, datatype.getName()));
             }
 
+            if ("curie".equals(datatype.getName()) && !fieldValue.textValue().contains(":")) {
```
karlbaker02: This is a tricky one because ideally this validation would be inside of the `isValid` method of the `CurieDatatype`, but you do not have access to the register field definitions in there as things stand.

src/test/java/uk/gov/register/service/ItemValidatorTest.java
```diff
@@ -44,28 +44,68 @@ public void setup() throws NoSuchConfigException, IOException {
...
+        when(countryRegisterMetadata.getFields()).thenReturn(Arrays.asList("country", "start-date", "curie-info", "curie-info2"));
 
         Field textField = new Field("text", "text", registerName, Cardinality.ONE, "text text");
         Field registerField = new Field("register", "text", registerName, Cardinality.ONE, "register text");
         Field fieldsField = new Field("fields", "string", registerName, Cardinality.MANY, "fields text");
 
         Field countryField = new Field("country", "string", countryRegisterName, Cardinality.ONE, "countryName");
-        Field startDateField = new Field("start-date\"", "datetime", null, Cardinality.ONE, "Start date text");
+        Field startDateField = new Field("start-date", "datetime", null, Cardinality.ONE, "Start date text");
+        final Field curieInfoField = new Field("curie-info", "curie", forCurieRegisterName, Cardinality.ONE, "Link to curie");
```
karlbaker02: Could we rename these two fields to `curieFieldWithRegisterSpecified` and `curieFieldWithoutRegisterSpecified` respectively please? (or names along those lines)

src/test/java/uk/gov/register/service/ItemValidatorTest.java
```diff
@@ -44,28 +44,68 @@ public void setup() throws NoSuchConfigException, IOException {
...
+
+        try {
+            countryItemValidator.validateItem(jsonNode, countryFieldsByName, countryRegisterMetadata);
+        } catch (ItemValidationException e) {
+            fail("Must not execute this statement");
+        }
+    }
+
+    @Test
+    public void validateItem_throwsValidationException_givenCueriWithoutColonNorRegister() throws IOException {
```
karlbaker02: Spelling mistake, but maybe something like: `validateItem_throwsValidationException_whenBothCurieAndFieldDefinitionDoNotSpecifyRegister`

karlbaker02:
One more comment: after testing, the curie validation is successful when you specify a curie of `:` or `ABC:`. These curies still cause issues loading data on the `/records` page.

Note that a curie of `:XYZ` doesn't cause any immediate problems rendering `/records`, but it will most likely cause problems further down the line as the `RegisterName` property won't be set properly for the curie.

PR #362, opened 2017-09-28
Force downloads in browsers for the different formats supported

fjavierm:


src/main/java/uk/gov/register/resources/ItemResource.java
```diff
@@ -18,11 +19,14 @@
 public class ItemResource {
     private final RegisterReadOnly register;
     private final ViewFactory viewFactory;
+    private final HeaderProvider headerProvider;
 
     @Inject
-    public ItemResource(RegisterReadOnly register, ViewFactory viewFactory) {
+    public ItemResource(RegisterReadOnly register, ViewFactory viewFactory,
+                        final RequestContext requestContext) {
```
gidsg: indentation is weird here?

fjavierm: Parameters are aligned, is it weird?

src/main/java/uk/gov/register/resources/EntryResource.java
```diff
@@ -52,7 +52,8 @@ public EntryResource(RegisterReadOnly register, ViewFactory viewFactory, Request
     @Timed
     public Optional<EntryListView> findByEntryNumber(@PathParam("entry-number") int entryNumber) {
         Optional<Entry> entry = register.getEntry(entryNumber);
-        return entry.isPresent() ? Optional.of(new EntryListView(Arrays.asList(entry.get()))) : Optional.empty();
+        headerProvider.setAttachmentContentDisposition(String.valueOf(entryNumber));
+        return entry.map(entry1 -> new EntryListView(Collections.singletonList(entry1)));
```
gidsg: it is just a style preference thing, but I would rather call this `e` than `entry1`.

fjavierm: Auto-generation, I didn't event read it. I'll change it.

src/main/java/uk/gov/register/resources/ItemResource.java
gidsg: OK, yes, that is fine, think I was just confused how it was looking on github,

src/main/java/uk/gov/register/resources/RecordResource.java
```diff
@@ -97,6 +98,9 @@ public EntryListView getAllEntriesOfARecord(@PathParam("record-key") String key)
     @Timed
     public RecordsView facetedRecords(@PathParam("key") String key, @PathParam("value") String value) {
         List<Record> records = register.max100RecordsFacetedByKeyValue(key, value);
+
+        headerProvider.setAttachmentContentDisposition(key);
+
```
gidsg: can remove whitespace here.

fjavierm: I'll do it, but I'm not a big fun of all the lines together, make it more difficult to read it

PR #365, opened 2017-10-05
Use HTML5 download parameter for download CTAs

gidsg:
Previously we were forcing download content disposition in all cases. Now we only set it where there is a specific download call to action.

Notes
* This relies on the [HTML5 `download` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download) which has [limited browser support](https://caniuse.com/#feat=download). 
* I am no longer setting the custom filenames which were triggered on download previously. So we rely on the default filename from the server e.g. `records.json` when the file is downloaded.


src/main/java/uk/gov/register/resources/EntryResource.java
```diff
@@ -59,8 +59,7 @@ public EntryResource(RegisterReadOnly register, ViewFactory viewFactory, Request
     @Timed
     public Optional<EntryListView> findByEntryNumber(@PathParam("entry-number") int entryNumber) {
         Optional<Entry> entry = register.getEntry(entryNumber);
-        headerProvider.setAttachmentContentDisposition(String.valueOf(entryNumber));
-        return entry.map(function -> new EntryListView(Collections.singletonList(function)));
+        return entry.isPresent() ? Optional.of(new EntryListView(Arrays.asList(entry.get()))) : Optional.empty();
```
fjavierm: Why this conditional structure? The map function does exactly this.

src/main/resources/templates/fragments/download-and-preview.html
```diff
@@ -5,16 +5,16 @@
...
+         data-download-type="spreadsheet" download="">spreadsheet</a>.
       <th:block th:if="${pagination.totalEntries} le 100">You can also preview as</th:block>
       <th:block th:if="${pagination.totalEntries} gt 100">You can also preview a sample as</th:block>
       <a th:href="${'/preview-' + records_or_entries_string + '/json'}">JSON</a>,
       <a th:href="${'/preview-' + records_or_entries_string + '/yaml'}">YAML</a> or
       <a th:href="${'/preview-' + records_or_entries_string + '/ttl'}">TTL</a>.
     </p>
   </div>
-</html>
+</html>
```
fjavierm: NIT: Not line at the end of the file

src/main/resources/templates/fragments/download-and-preview-home.html
```diff
@@ -4,9 +4,9 @@
     <p class="data-representations">
       <th:block th:if="${totalRecords} le 5000">Download a copy of this register as</th:block>
       <th:block th:if="${totalRecords} gt 5000">Download the first 5,000 records as</th:block>
-      <a href="/records.csv?page-size=5000" rel="alternate" class="js-download" data-download-type="CSV">CSV</a>,
-      <a href="/records.tsv?page-size=5000" rel="alternate" class="js-download" data-download-type="TSV">TSV</a> or as a
-      <a href="/records.xlsx?page-size=5000" rel="alternate" class="js-download" data-download-type="XLSX">spreadsheet</a>.
+      <a href="/records.csv?page-size=5000" rel="alternate" class="js-download" data-download-type="CSV" download="">CSV</a>,
```
fjavierm: @danmitchell- can correct me on this but as far as I know, Custom Data Attributes start with `data-` in HTML5.

src/main/java/uk/gov/register/resources/RecordResource.java
```diff
@@ -20,18 +20,18 @@
 @Path("/")
 public class RecordResource {
     private final HttpServletResponseAdapter httpServletResponseAdapter;
+    private final RequestContext requestContext;
```
fjavierm: `requestContext` is only used in the constructor, why are you adding it as property?

src/main/java/uk/gov/register/resources/RecordResource.java
```diff
@@ -20,18 +20,18 @@
...
     private final RegisterReadOnly register;
     private final ViewFactory viewFactory;
     private final RegisterName registerPrimaryKey;
```
fjavierm: `registerPrimaryKey` is not used any more, except in the constructor, can it be removed?

src/main/resources/templates/fragments/download-and-preview-home.html
gidsg: @danmitchell- this is not a data attribute. See https://www.w3schools.com/tags/att_a_download.asp

fjavierm: @gidsg I saw that, but I thought it was just `download` without `=""` . I was not sure what it was. Just ignore the comment then

danmitchell-: Correct. Its not a `data` attribute.

gidsg:
@fjavierm I think once you are doing an `Open in New Tab` generally the browser is treating that as a new request so ignores the `download` attribute. I am not sure how big an issue this is.

fjavierm:
@gidsg My point is that with the `ContentDisposition` attachement, this was working as expected.

gidsg:
@fjavierm I am not sure if the user selects Open In New Tab what exactly they are expecting. If this is really critical we could implement it server-side as a query string parameter which sets the `ContentDisposition` but I would argue the intended behaviour is ambiguous in this case. Happy to talk with you about this.

src/main/resources/templates/fragments/download-and-preview-home.html
danmitchell-: Not sure if im missing something here but you should be passing the URL into the download attribute i.e. `<a href="/preview-records/json" download="/preview-records/json">JSON</a>`

gidsg: No, it uses the URL from the href. The value is optional and is used to override the filename given by the server.

danmitchell-: Ah yes. Not sure how much I trust that but 👌

PR #366, opened 2017-10-06
Validate consecutive append entries

karlbaker02:
This PR adds validation to ensure that identical `append-entry` commands result in a validation error when calling `load-rsf`. The reason for this is because we do not want to append an entry to a register if there is no change from the item(s) assigned against the previous entry for the same key. Doing this ensures that any consumers of the API do not write duplicate data to registers.


src/test/java/uk/gov/register/indexer/IndexDriverTest.java
```diff
@@ -360,7 +360,7 @@ public void indexEntry_shouldStartIndexSpecifyingStartIndexEntryNumber_whenItemE
         verify(dataAccessLayer, times(0)).end(anyString(), anyString(), anyString(), anyString(), anyInt(), anyInt(), anyInt());
     }
 
-    @Test
+//    @Test
```
karlbaker02: Need to check that this method is no longer needed.

fjavierm:
@karlbaker02 I've rebased the PR

karlbaker02:
@fjavierm We should put some detail in the commit message - I'll do that in a bit.

michaelabenyohai:
@karlbaker02 Please can we add some detail to the commit before merging?

karlbaker02:
Will do @michaelabenyohai. Did you check the changes I made to the migration by the way?

PR #368, opened 2017-10-16
Accessibility fixes

karlbaker02:
This commit makes two accessibility fixes: to change the 'Using this register' heading on each register homepage from a `h3` to a `h2`, and to remove the link from the OGL logo in the page footer.

**Heading change:**

Before:

screen shot 2017-10-16 at 12 03 12

After:

screen shot 2017-10-16 at 12 05 20

**OGL logo link change:**

Before:

screen shot 2017-10-16 at 12 06 26

After:

screen shot 2017-10-16 at 12 12 52

### DIFF
--- a/src/main/resources/assets/sass/main.scss
+++ b/src/main/resources/assets/sass/main.scss
@@ -315,3 +315,13 @@ dl {
 pre {
     margin: 0;
 }
+
+.open-government-licence {
+  .logo {
+    background: image-url("images/open-government-licence.png") 0 0 no-repeat;
+	@include device-pixel-ratio() {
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/sass/main.scss
```diff
@@ -315,3 +315,13 @@ dl {
 pre {
     margin: 0;
 }
+
+.open-government-licence {
+	.logo {
```
danmitchell-: Indentation looks odd here. Should be 2 spaces

karlbaker02: Addressed.

PR #372, opened 2017-10-27
add charity and charity-class to discovery

gidsg:


gidsg:
close in favour of #371

PR #376, opened 2017-11-06
Call getEntriesByKeys for each key batch

karlbaker02:
This PR changes the invocation of `getEntriesByKeys` to pass in the entry keys that have been generated in batches of 1000, as opposed to passing in all of the entry keys (as this defeats the purpose of batching for performace reasons).


michaelabenyohai:
Tested and resolves the specific issue I faced loading charity data 👍

PR #377, opened 2017-11-06
Add the-electoral-commission to public-bodies

karlbaker02:
This PR adds `the-electoral-commission` to `public-bodies.yaml`, as it is defined as the registry by the new `westminster-parliamentary-constituency` register.


michaelabenyohai:
Approved, assuming it builds.

PR #383, opened 2017-11-15
Fix run-application.sh script

arnau:
Context: `run-application.sh` failed to run in a clean enviornment due
to missing dependencies stored in git submodules.

Solution: Now `run-application.sh` initialises _and_ updates git
submodules before doing any other work.

### DIFF
--- a/run-application.sh
+++ b/run-application.sh
@@ -36,6 +36,8 @@ function do_nothing_forever {
 
 trap on_exit EXIT
 
+git submodule update --init
+
 if [ ! -e "./deploy/openregister-java.jar" ]
 then
   docker run \


arnau:
@michaelabenyohai the `RUNNING-LOCALLY.md` file suggests twice to have a look at the configuration to know more about options. Where is this configuration placed?

michaelabenyohai:
@arnau it lives at `config.docker.register.yaml` and `config.basic.register.yaml`, mentioned in the "Run register locally" section. The "clone a register" and "create a new register" then tell you how to change the configuration for specific things you might like to do.

arnau:
If I follow the steps for loading `school-type` instead of `country`, I get either a `Register undefined` or an exception `curl: (22) The requested URL returned error: 400 Bad Request` depending on the `ENVIRONMENT` value. Thoughts?

arnau:
Ok, I will open a new issue for the above so I can merge this PR.

PR #385, opened 2017-11-16
Add code of conduct 1.4

arnau:
Adds a default code of conduct 1.4 with email `registersteam+code-of-conduct@digital.cabinet-office.gov.uk`.

Let me know if we want to get any inspiration from https://github.com/alphagov/code-of-conduct

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,6 +1,21 @@
-# Contributor Covenant Code of Conduct
+# Code of conduct for `openregister`
 
-## Our Pledge
+Contributors to repositories hosted in `openregister` are expected to
+follow the Contributor Covenant Code of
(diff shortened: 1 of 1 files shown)


michaelabenyohai:
@arnau I'm wondering whether we should take the alphagov Code of Conduct as is and adapt to openregister?

arnau:
If we move to alphagov we need eventually to use theirs as far as I understand. That would mean we don't have an email in there pointing to us. I'm fine either way.

arnau:
@michaelabenyohai @elliecraven as discussed with Ellie and Andy, I changed it to be an adapted version of alphagov.

danmitchell-:
No objections from me 👍

elliecraven:
Looks good to me too 👍

arnau:
As discussed, going ahead and merge. We will evolve it when alphagov does it.

PR #391, opened 2017-11-23
Rotate secret key for deployment

michaelabenyohai:
On deployment we push artifacts to S3 buckets. This rotates the
key used to access S3.


michaelabenyohai:
I haven't been able to find a way of verifying it unfortunately. I would have liked to be able to do that.

PR #394, opened 2017-11-28
Add github templates

arnau:
Same github templates we have in the management tool repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,5 @@
+### What did you do? (steps to reproduce)
+
+### What happened? (actual results)
+
+### What should have happened? (expected results)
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+### What did you do? (steps to reproduce)
```
gidsg: I would add a heading above here 'Summary' and at the bottom 'Environment Info/URLs'

arnau: Is this something you would like to have here but not in the RMT or should we align the two?

arnau: Ok, as agreed during the standup I'll merge it as is and we can iterate on it later on.

PR #410, opened 2018-01-03
Use existing DBIHealthCheck

karlbaker02:
### Context

This PR creates a new `healthcheck` endpoint which CloudFoundry will be able to use to determine whether the application is healthy. The new endpoint uses the existing `openregister_java` health check created by JDBI when a new `DBI` instance is created as part of the `DatabaseManager`.

When the `openregister_java` health check runs, it attempts to connect to the backing database within 5 seconds and run a `SELECT 1` query - if this is unsuccessful then the application is deemed unhealthy.

A number of options were considered for exposing the health checks that are available within the admin interface of DropWizard:

The first was to simply have CloudFoundry connect to the admin interface via the separate port, however this is not supported by CloudFoundry - it can only connect to the app's default port if running a HTTP healthcheck (https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html#-health-check-http-endpoints).

The next option considered was to host the admin interface of DropWizard on the same port as the default app, to circumvent the limitations of CloudFoundry, but this has the downside that all parts of the admin interface are exposed publicly. This interface can be password protected but would require CloudFoundry to be able to support this in order to consider the application's health. Intercepting requests to the admin interface and redirecting all to the healthcheck page via a custom filter was also explored, but it was not possible to intercept these requests.

Another option considered was to implement a custom filter to detect any requests to `/healthcheck` and run the health check at this point, however somewhere there exists some middleware which detects that the endpoint does not exist and changes the response accordingly.

The final option explored, which is the contents of this commit, creates a new `HealthcheckResource` and returns a response based on the health reported by the health check. In particular, we register the result of `environment.healthChecks()` to `HealthCheckRegistry.class` in `RegisterApplication` to be able to inject this directly into our resource without the need for the whole environment.

Add new `HealthcheckResource`.

### DIFF
--- a/deploy/manifests/alpha/basic.yml
+++ b/deploy/manifests/alpha/basic.yml
@@ -5,7 +5,7 @@ applications:
   instances: 2
   buildpack: java_buildpack
   health-check-type: http
-  health-check-http-endpoint: /records
+  health-check-http-endpoint: /healthcheck
   path: ../../openregister-java.jar
   domains:
     - alpha.openregister.org
--- a/deploy/manifests/alpha/multi.yml
+++ b/deploy/manifests/alpha/multi.yml
@@ -5,7 +5,7 @@ applications:
   instances: 2
   buildpack: java_buildpack
   health-check-type: http
-  health-check-http-endpoint: /records
+  health-check-http-endpoint: /healthcheck
   path: ../../openregister-java.jar
   domains:
     - alpha.openregister.org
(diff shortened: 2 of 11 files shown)


karlbaker02:
Updated @gidsg

gidsg:
we should probably also update the example in https://github.com/openregister/openregister-java/blob/master/manifest.yml

michaelabenyohai:
👍

PR #412, opened 2018-01-15
Update ga tracking

danmitchell-:
### Context
We want a consistent approach across all GaaP products for tracking so we're using https://github.com/alphagov/gaap-analytics to achieve this.

Update Google Analytics tracking and move init into head as recommended.


src/main/resources/assets/js/analytics-tracking.js
```diff
@@ -0,0 +1,67 @@
+// Works in combination with the following data-attributes
```
michaelabenyohai: This file seems to be a copy of the code at https://github.com/alphagov/gaap-analytics/tree/master/src/. Can we add those scripts as a dependency rather than copying them into our project, so that we can take advantages of any updates to it? I'm probably missing something here.

danmitchell-: Correct, it’s a compiled version of this. We can’t use this out of the box because it’s written in ES6 so we’d need Babel to compile to ES5.

michaelabenyohai: As discussed, given that we cannot include this as a dependency can we add a comment to the file explaining that this code comes from https://github.com/alphagov/gaap-analytics/tree/master/src/. We should consider including this dependency properly if/when we come to refactoring the frontend.

michaelabenyohai:
All good apart from the small comment above

michaelabenyohai:
@danmitchell- Travis is having some problems with the git submodule.

PR #417, opened 2018-01-17
Remove history page url from config and views

danmitchell-:
### Context
The history no longer exists so there is no need to retain the functionality

Remove anything relating to `historyPageUrl`

### Guidance to review
Ensure the home page still renders and all tests pass


michaelabenyohai:
We already removed them all when we decided history pages weren't a thing. But there is some redundant code that we can remove - I'll add a card for it.

PR #418, opened 2018-01-17
Remove old discovery registers

danmitchell-:
No longer needed or promoted have been promoted up.

cc @elliecraven

### DIFF
--- a/deploy/manifests/discovery/multi.yml
+++ b/deploy/manifests/discovery/multi.yml
@@ -11,22 +11,15 @@ applications:
     - discovery.openregister.org
   hosts:
     - address
-    - approved-meat-establishment
     - charity
     - charity-class
     - clinical-commissioning-group
-    - green-deal-certification-body
-    - meat-establishment-operation
-    - meat-establishment-outcome
-    - meat-establishment-type
     - place
     - police-force
     - police-neighbourhood
-    - prison
     - street-custodian
     - street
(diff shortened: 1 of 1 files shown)


elliecraven:
@danmitchell- looks fine to me to remove all of these, but worth checking with @nacnudus whether any of the meat ones need to stay for us to pick up later, or if we're starting from scratch with those?

nacnudus:
@elliecraven Dan already did :) I think they can be deleted because they'll probably be replaced by more-general 'food' registers, which all follow the same structure.

michaelabenyohai:
Why are we removing street-custodian when it is linked to from street and address @danmitchell- @nacnudus ?

michaelabenyohai:
We also need a corresponding PR for the change in the `deployment` repository. If we merge this the register endpoints will still exist but we'll simply get a PaaS error returned.

nacnudus:
`street` and `address` can also be deleted @danmitchell- @michaelabenyohai from my point of view.

danmitchell-:
I've left `street-custodian` as it links with `street` and `address`. We can remove them all separately if we wish.

michaelabenyohai:
I have just noticed that `place` shouldn't be removed because it is linked to from `street`.

danmitchell-:
Also put `place` back on this one. Please approve and i'll merge

PR #438, opened 2018-02-16
Strip back HTML views

danmitchell-:
### Context
Our new frontend is the intended entry point for registers so all the contextual information relating to them has been moved to there.

Remove "fluff"

### Guidance to review
Check all the views.


danmitchell-:
Can/Should remove `custodianName` from all view resources too

danmitchell-:
Can/Should remove `lastUpdated` from all view resources too

danmitchell-:
Can/Should remove `registerLinkService` from all view resources too

danmitchell-:
Can remove alot more methods from this file `src/main/java/uk/gov/register/views/HomePageView.java`

michaelabenyohai:
A

src/main/resources/templates/home.html
```diff
@@ -11,78 +11,40 @@
...
-          <p>The <span th:utext="${friendlyRegisterName}"></span> contains <span th:text="${totalRecords}"></span>
-            records and includes the following fields:</p>
-          <div th:replace="fragments/field-list.html::field-list(${fields})"></div>
+          </span>
+
+          <h2 class="heading-medium">Using this register</h2>
 
-          <h2>Using this register</h2>
+          <h3 class="heading-small">Base URL</h3>
+          <pre class="preview-box"><code class="language-markup" th:utext="${registerResolver.baseUriFor(registerName)}"></code></pre>
```
karlbaker02: This gives me the following when running `country` locally:

screen shot 2018-02-21 at 12 05 19

src/main/resources/templates/home.html
```diff
@@ -11,78 +11,40 @@
...
-                                                                target="_blank">technical documentation</a>.</p>
-          <p>If you're building a service, you can use this data to create elements for forms.</p>
-          <p>You might need to combine data from this register with data from another source to do this.</p>
-        </main>
+
+          <p><code class="code">/records</code> returns the most recent item for every key in a register.</p>
+
+          <p><code class="code">/entries</code> returns the log of all entries appended to a register and the hash of the item attached to each entry, ordered by entry number.</p>
+
+          <p><code class="code">/item{hash}</code> returns an item is a blob of data. Each item in a register is linked to an entry by a unique hash. The entry describes when the item was added to the register.</p>
```
karlbaker02: Should be _returns an item **as** a blob of data_.

Also, the endpoint should be `/item/{hash}`, and an example might be useful to specify that the prefix will be `sha-256:`

src/main/resources/templates/fragments/download-and-preview-single.html
```diff
@@ -2,17 +2,14 @@
 <html xmlns:th="http://www.thymeleaf.org" xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
   <div th:fragment="download-and-preview-single(records_or_entries_or_items_string)">
     <p class="data-representations">
-      Download a copy as
+      Get this data as
+      <a th:href="${'/' + records_or_entries_or_items_string + '.json/' + (#strings.equals(records_or_entries_or_items_string, 'records') ? content.entryKey : (#strings.equals(records_or_entries_or_items_string, 'entries') ? content.entryNumber : content.itemHash))}">JSON</a>,
+      <a th:href="${'/' + records_or_entries_or_items_string + '.yaml/' + (#strings.equals(records_or_entries_or_items_string, 'records') ? content.entryKey : (#strings.equals(records_or_entries_or_items_string, 'entries') ? content.entryNumber : content.itemHash))}">YAML</a>,
```
karlbaker02: YAML no longer working (tested on the Item page).

src/main/resources/templates/fragments/download-and-preview-single.html
```diff
@@ -2,17 +2,14 @@
...
+        <a th:href="${'/' + records_or_entries_or_items_string + '.ttl/' + (#strings.equals(records_or_entries_or_items_string, 'records') ? content.entryKey : (#strings.equals(records_or_entries_or_items_string, 'entries') ? content.entryNumber : content.itemHash))}">TTL</a>
```
karlbaker02: TTL no longer working (tested on the Item page).

karlbaker02:
Have we made a decision to remove breadcrumbs from the API inspector?

src/main/resources/templates/home.html
danmitchell-: I get the same but assumed it was being it was local?

src/main/resources/templates/home.html
danmitchell-: This text is being reviewed by @michaelabenyohai

src/main/resources/templates/fragments/download-and-preview-single.html
danmitchell-: Looking

src/main/java/uk/gov/register/views/HomePageView.java
```diff
@@ -51,7 +51,7 @@ public HomePageView(
 
     @SuppressWarnings("unused, used from template")
     public String getRegisterText() {
-        return markdownProcessor.markdown(getRegister().getText());
```
michaelabenyohai: I think we should leave this here? We still want to render `text` fields as markdown if we can.

src/main/resources/templates/home.html
michaelabenyohai: yes agree I think it's a problem unfortunately

src/main/resources/templates/fragments/download-and-preview.html
```diff
@@ -2,21 +2,17 @@
...
+      <a th:href="${'/' + records_or_entries_string + '.json'}">JSON</a>,
+      <a th:href="${'/' + records_or_entries_string + '.yaml'}">YAML</a>,
+      <a th:href="${'/' + records_or_entries_string + '.ttl'}">TTL</a>,
       <span data-click-events="" data-click-category="Content" data-click-action="Register Download">
-        <th:block th:if="${pagination.totalEntries} le 5000">Download a copy of this register as</th:block>
-        <th:block th:if="${pagination.totalEntries} gt 5000">Download the first 5,000 records as</th:block>
         <a th:href="${'/' + records_or_entries_string + '.csv?page-size=5000'}" rel="alternate" class="js-download"
-          data-download-type="CSV" download="">CSV</a>,
+          data-download-type="CSV" download="">CSV</a> or
         <a th:href="${'/' + records_or_entries_string + '.tsv?page-size=5000'}" rel="alternate" class="js-download"
```
michaelabenyohai: I think we need to remove the page-size for CSV and TSV here to make it consistent with the other formats?

src/main/java/uk/gov/register/views/HomePageView.java
danmitchell-: We only expect a string in the frontend so that we can control the styling better. Why do we want markdown?

src/main/resources/templates/entries.html
```diff
@@ -22,7 +22,7 @@ <h1 class="heading-large">
             <span class="heading-secondary"
                   th:include="fragments/pagination.html::pagination-top('entries')"></span>
           </h1>
-          <div th:include="fragments/download-and-preview.html::download-and-preview('entries')"></div>
+          <div th:include="fragments/download-and-preview.html::download-and-preview(${'record/' + content.recordKey.get() + '/entries'})"></div>
```
michaelabenyohai: I believe this should be `/entries` and if this is being used to show the /entries page. This is currently returning a 500 and failing tests @karlbaker02

src/main/java/uk/gov/register/views/HomePageView.java
michaelabenyohai: It's only for fields that are of datatype `text` because we specify in the datatype register that that datatype can contain markdown https://datatype.register.gov.uk/record/text. However I think this might be broken in prod currently anyway 😕

src/main/resources/templates/entries.html
karlbaker02: This makes sense. I changed this earlier on because downloads of entries for individual keys downloaded _all_ entries, not just what is being viewed (FYI this is also the case in production, so an existing bug).

karlbaker02: I will revert this for now and we can fix later on.

PR #439, opened 2018-02-19
Update DCLG to Ministry of Housing, Communities & Local Government

karlbaker02:
### Context
This PR updates _Department for Communities and Local Government_ to _Ministry of Housing, Communities & Local Government_.


karlbaker02:
Ah yes, sorry you're right. I'll add this as an additional entry for now.

PR #443, opened 2018-02-22
Add fire authority alpha environment

danmitchell-:
FYI @trvllngmn


deploy/manifests/alpha/multi.yml
```diff
@@ -23,6 +23,8 @@ applications:
     - social-housing-provider-legal-entity-eng
     - social-housing-provider-eng
     - statistical-geography-council-area-sct
+    - statistical-geography-council-area-sct
```
michaelabenyohai: You've accidentally added this one twice

PR #444, opened 2018-02-22
Use `apt-get clean` in `buildspec.yml`

michaelabenyohai:
### Context
This should hopefully fix the issue in code deploy where it cannot perform `apt-get upgrade` as per https://askubuntu.com/questions/41605/trouble-downloading-packages-list-due-to-a-hash-sum-mismatch-error

Use `apt-get clean` in `buildspec.yml`

### Guidance to review
We will have to see if it works when we deploy.

### DIFF
--- a/deploy/buildspec.yml
+++ b/deploy/buildspec.yml
@@ -3,6 +3,7 @@ version: 0.2
 phases:
   install:
     commands:
+      - apt-get clean -y
       - apt-get update -y
       - apt-get install -y apt-transport-https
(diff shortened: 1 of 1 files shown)


deploy/buildspec.yml
```diff
@@ -3,6 +3,7 @@ version: 0.2
 phases:
   install:
     commands:
+      - apt-get clean -y
```
arnau: If it doesn't solve the problem entirely. Perhaps you can explore the `--fix-broken` flag. See https://linux.die.net/man/8/apt-get

PR #451, opened 2018-02-26
Fix download link for entries of a record

karlbaker02:
### Context

This PR fixes the location of downloads for all formats when a user attempts to download the entries for a particular key. Previously, the download would point to `/entries` for the particular format, meaning that all entries would be downloaded, even when the user would be viewing only entries for some key. This PR now points downloads of this type to `/record/KEY/entries` for the particular format.

Fix download link for entries of a record.

### Guidance to review

Start ORJ, go to any record, view all versions of this record and then try downloading the resulting entries via the different formats.

### DIFF
--- a/src/main/java/uk/gov/register/views/EntryListView.java
+++ b/src/main/java/uk/gov/register/views/EntryListView.java
@@ -95,6 +95,12 @@ public class EntryListView implements CsvRepresentationView {
         return recordKey;
     }
 
+    @SuppressWarnings("unused, used from templates")
+    public String getDownloadLink() {
+        Optional<String> recordKey = getRecordKey();
+        return recordKey.isPresent() ? String.format("record/%s/entries", recordKey.get()) : "entries"; 
+    }
+
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/views/EntryListView.java
```diff
@@ -95,6 +95,12 @@ public String entriesTo(final String mediaType, final Provider<RegisterName> reg
         return recordKey;
     }
 
+    @SuppressWarnings("unused, used from templates")
+    public String getDownloadLink() {
+        Optional<String> recordKey = getRecordKey();
+        return recordKey.isPresent() ? String.format("record/%s/entries", recordKey.get()) : "entries";
```
gidsg: could you just call `getRecordKey()` directly here?

karlbaker02: I can but then that's two calls rather than one, which is why I didn't. Don't mind either way though.

gidsg: OK makes sense, happy to leave it as-is!

PR #452, opened 2018-02-26
Download should use correct page info

karlbaker02:
### Context

This PR fixes a bug for both `/records` and `/entries` which currently leads you to download the wrong data, unless you're viewing page 1 with the default page size of 100 (or for entries, the default `start` of 1 and `limit` of 100).

With this PR, a user wishing to download the particular `/records` they're viewing, say page=3 with page-size=150, will be able to download exactly this data. Similarly, they can download the `/entries` they're viewing, from start=150 with limit=200.

Fix download links for `/records` and `/entries` to include paging info.

### Guidance to review

Go to `/records`, change `page-index` and `page-size`, then inspect the download URLs and try to download in various formats.

Similarly, for `/entries`, change `start` and `limit`, then inspect the download URLs and try to download in various formats.

### DIFF
--- a/src/main/java/uk/gov/register/resources/IndexSizePagination.java
+++ b/src/main/java/uk/gov/register/resources/IndexSizePagination.java
@@ -91,9 +91,13 @@ public class IndexSizePagination implements Pagination {
     public String getPreviousPageLink() {
         return String.format("?" + INDEX_PARAM + "=%s&" + SIZE_PARAM + "=%s", getPreviousPageNumber(), pageSize());
     }
-
+    
     public int pageSize() {
         return pageSize;
+    }
+    
+    public String getFormattedPageQueryParams() {
+        return String.format("?%s=%s&%s=%s", INDEX_PARAM, pageIndex, SIZE_PARAM, pageSize);
     }
 
     @Override
--- a/src/main/java/uk/gov/register/resources/Pagination.java
+++ b/src/main/java/uk/gov/register/resources/Pagination.java
@@ -22,4 +22,6 @@ public interface Pagination {
     String getPreviousPageLink();
 
     int getTotalPages();
+
+    String getFormattedPageQueryParams();
 }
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/register/resources/IndexSizePagination.java
```diff
@@ -91,10 +91,14 @@ public String getNextPageLink() {
...
         return String.format("?" + INDEX_PARAM + "=%s&" + SIZE_PARAM + "=%s", getPreviousPageNumber(), pageSize());
     }
-
+    
     public int pageSize() {
         return pageSize;
     }
+    
+    public String getFormattedPageQueryParams() {
+        return String.format("?%s=%s&%s=%s", INDEX_PARAM, pageIndex, SIZE_PARAM, pageSize);
```
gidsg: Is there a Java library for generating query strings rather than doing it by hand?

karlbaker02: Not sure, although we do use `String.format` in other places in the codebase.

src/main/java/uk/gov/register/resources/IndexSizePagination.java
```diff
@@ -91,10 +91,14 @@ public String getNextPageLink() {
     public String getPreviousPageLink() {
         return String.format("?" + INDEX_PARAM + "=%s&" + SIZE_PARAM + "=%s", getPreviousPageNumber(), pageSize());
     }
-
+
```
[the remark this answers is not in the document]
karlbaker02: Let me double check that.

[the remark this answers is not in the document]
karlbaker02: I can't see anything amiss in the actual code.

src/main/java/uk/gov/register/resources/IndexSizePagination.java
karlbaker02: Looking at this more, there are things out there such as `UriBuilder` to build an entire URI, though I'm not sure how applicable this is for the current situation, given that these URIs are all generated within the template files.

PR #456, opened 2018-03-22
 Fix bug where URL datatypes cause TTL views to return 500

michaelabenyohai:
### Context
Currently fields with URL datatype cause TTL endpoints to 500. This is due to the below error.

```
"ERROR","level_value":40000,"stack_trace":"java.lang.ClassCastException: uk.gov.register.core.UrlValue cannot be cast to uk.gov.register.core.LinkValue\n\tat uk.gov.register.views.representations.turtle.ItemTurtleWriter$FieldRenderer.renderScalar(ItemTurtleWriter.java:77)
```

This prevents a `ClassCastException` occurring in `ItemTurtleWriter`, which was previously trying to cast a `UrlValue` to a `LinkValue`. This change renames `LinkValue` to `RegisterLinkValue`. Both `UrlValue` and `RegisterLinkValue` now extend the new `LinkValue` class. The UrlTemplateLinkResolver` now resolves `LinkValues`, which prevents casting problems.

### Guidance to review
Ensure /records.ttl and /item/sha-256:abcde.ttl return a 200 for registers that contain a URL datatype (e.g. government-organisation and local-authority-nir). URLs should also be expressed as resources rather than properties. I.e.

Resource:
```
<https://www.gov.uk>
```

Property:
```
"https://www.gov.uk"
```

### DIFF
--- a/src/main/java/uk/gov/register/core/RegisterLinkValue.java
+++ b/src/main/java/uk/gov/register/core/RegisterLinkValue.java
@@ -1,0 +1,51 @@
+package uk.gov.register.core;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * A value representing a link. Its "value" is a string representation. It also knows
+ * its target register, and its primary key within the target register.
+ *
+ * For a regular RegisterLinkValue, the "value" and "linkKey" are the same. For a CurieValue, the "value"
+ * is the Curie as a string, while the "linkKey" is the second half of the Curie (after the colon).
+ */
+public class RegisterLinkValue extends LinkValue implements FieldValue {
+    private final RegisterName targetRegister;
+    private final String value;
+    private final String linkKey;
+
+    public RegisterLinkValue(RegisterName registerName, String value) {
+        this(registerName, value, value);
+    }
+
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/register/core/RegisterLinkValue.java
```diff
@@ -0,0 +1,51 @@
...
+        return targetRegister;
+    }
+
+    public String getLinkKey() {
+        return linkKey;
+    }
+
+    public static class CurieValue extends RegisterLinkValue {
+        public CurieValue(String curieValue) {
+            super(new RegisterName(curieValue.split(":")[0]), curieValue, curieValue.split(":")[1]);
```
gidsg: very minor but I think you only need to do `curieValue.split(":")` once then you can read the two results separately?

michaelabenyohai: This code already existed but looks new as I renamed the class. Totally agree though so will change that.

michaelabenyohai: @gidsg just realised why it wasn't done this way - it's a constructor which means that `super()` has to be the first statement. Might have to leave it this way.

PR #457, opened 2018-03-22
Rename DCMS to DDCMS

nacnudus:
### Context

"Department for Culture, Media & Sport" is now "Department for Digital, Culture, Media & Sport" and has been renamed here https://www.gov.uk/api/organisations

This might be used as the registry of the data-sharing-agreement registers.  It hasn't yet been used as a registry of any other register.

Rename and re-slug "Department for Culture, Media & Sport" `department-for-culture-media-sport` to "Department for Digital, Culture, Media & Sport" `department-for-digital-culture-media-sport`.

### Guidance to review

Is it okay to rename these things, or should a new entry be appended instead?

### DIFF
--- a/src/main/resources/config/public-bodies.yaml
+++ b/src/main/resources/config/public-bodies.yaml
@@ -50,9 +50,9 @@ legal-services-commission:
 civil-service-intelligence-analysis-profession:
   name: "Intelligence Analysis "
   public-body: "civil-service-intelligence-analysis-profession"
-department-for-culture-media-sport:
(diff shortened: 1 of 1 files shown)


michaelabenyohai:
@nacnudus this is ok to do given that no existing registers have their `registry` as the old `department-for-culture-media-sport`. I've just checked all the Register registers and looks fine. It would have been breaking for those registers anyway as it's been removed from the https://www.gov.uk/api/organisations API.

PR #459, opened 2018-04-04
Separate register name from ID

michaelabenyohai:
### Context
Some registers will now use "meaningless" IDs. We also need a more human-readable register name. The register name will be optional as a first step. If it's not specified, the ID will continue to be used as a name.

See individual commits. There is also some refactor to the record/index layers to tidy up a bit. The commit messages should describe each change in more detail

### Guidance to review
If you add an entry against key `register-name` then you should see this name used on the home page. Otherwise the ID is used.

```
add-item  {"register-name":"School England"}
append-entry  system  register-name  ts sha-256:abc
```

TODO: rename this key to name and name to id at some point.


michaelabenyohai:
This is failing because of the vulnerability check. I'm going to fix that in another PR now.

karlbaker02:
Everything looks good (so many changes)! There'll have to be some work to remove some of the methods we have at the point in time when the entry tables are merged, but making it clearer in the meantime is probably a good idea. I really like that `DerivationRecordIndex` has been removed too - everything with the name `derivation` needs to go :)

PR #461, opened 2018-04-06
Refactor download-and-preview views

michaelabenyohai:
### Context
A refactor of download-and-preview HTML views.

We could have been using the `HtmlViewSupport` class for all download-and-preview HTML views, which simplifies the template logic. It also means we can remove `download-and-preview-single.html` as it does the same thing as `download-and-preview.html`.

### Guidance to review
Check all the `Get data as` links still work on all views for entry, item, record and including pagination.


michaelabenyohai:
Blocked by merging https://github.com/openregister/openregister-java/pull/460

gidsg:
are we still linking to these views? I can't find them in the current design of the API inspector?

michaelabenyohai:
@gidsg the pages for "preview json" etc no longer exist. However, these views are just reusable fragments for the "Get this data as JSON, YAML etc." HTML on most pages. So they are used in `entries.html`, `entry.html`, `records.html` etc.

PR #462, opened 2018-04-06
Make "View Register" lower case "View register"

michaelabenyohai:
### Context
This was picked up by Mark.


michaelabenyohai:
Blocked by merging https://github.com/openregister/openregister-java/pull/460

PR #463, opened 2018-04-09
Instructions update

maxf:
### Context

New user getting set up

- make `run-application.sh` script usable even if you use an incompatible shell
- use the correct classpath of module (`openregister-java` without `_main`)
- be agnostic about the pyvenv version (in case you have 3.6 for instance)

### Guidance to review

The last 2 points are open to discussions, but they're changes I've had to make to get set up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ See detailed steps for experimenting locally [here](RUNNING_LOCALLY.md).
 - `File > Open` your checkout directory
 - `Run > Edit Configurations`
 - Edit `RegisterRunner`
-- Use classpath of module: `openregister-java_main`
+- Use classpath of module: `openregister-java`
 - JRE: *select your installed JRE e.g.* `1.8`
 - `Run > Run... > RegisterRunner`
 
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ See detailed steps for experimenting locally [here](RUNNING_LOCALLY.md).
 - `File > Open` your checkout directory
 - `Run > Edit Configurations`
 - Edit `RegisterRunner`
-- Use classpath of module: `openregister-java`
+- Use classpath of module: `openregister-java_main`
 - JRE: *select your installed JRE e.g.* `1.8`
 - `Run > Run... > RegisterRunner`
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -161,7 +161,7 @@ task(loadSchoolDataForConformance, type: Exec) {
 }
 
 task createConformanceTestVenv(type: Exec) {
-    commandLine 'pyvenv-3.5', '.venv'
+    commandLine 'pyvenv', '.venv'
```
arnau: Does this ensure python3 is used? Or, does this prevent using python2?

maxf: Looks like pyvenv is deprecated anyway: https://docs.python.org/3/library/venv.html
It says we should be using `python -m venv`, which doesn't seem to work with python 2.

Would mandating python3 be acceptable?

arnau: I think it is acceptable, yes. @gidsg @michaelabenyohai ?

gidsg: @maxf you can try this although we have had problems with this failing on Travis CI when we attempted to switch from `pyvenv` currently we have this condition to use an old travis build which does not come bundled with Python 3: https://github.com/openregister/openregister-java/blob/086128b2b0ea8fd07db8d5dd2dcdecc410c60ffe/.travis.yml#L1

My suggestion would be to merge this as-is as it is an improvement on what we have and then in a separate PR look to update the travis build and use python3 `venv`.

README.md
```diff
@@ -56,7 +56,7 @@ See detailed steps for experimenting locally [here](RUNNING_LOCALLY.md).
 - `File > Open` your checkout directory
 - `Run > Edit Configurations`
 - Edit `RegisterRunner`
-- Use classpath of module: `openregister-java_main`
+- Use classpath of module: `openregister-java`
```
gidsg: not sure about this, if I use this classpath I get error:
```
Could not find or load main class uk.gov.register.RegisterApplication
```

maxf: I'll remove this for now then, and will investigate why it's the other way round for me.

build.gradle
maxf: fine by me!

PR #465, opened 2018-04-23
Implement RFC 0002 (URI Consolidation)

gidsg:
### Context
https://github.com/openregister/registers-rfcs/blob/master/content/uri-consolidation/index.md

Implement [RFC 0002](https://github.com/openregister/registers-rfcs/blob/master/content/uri-consolidation/index.md)

### Guidance to review
Canonical URLs should be changed as per RFC e.g. http://localhost:8080/records/400673.json should resolve
Old URLs should 301 redirect as per RFC e.g. http://localhost:8080/record/400673 --> http://localhost:8080/records/400673

N.B. Conformance tests are broken because they do not reflect this RFC. Specifically the URLs within the Turtle representation. We should update them after this PR is merged.

### DIFF
--- a/src/main/java/uk/gov/register/resources/VerifiableLogResource.java
+++ b/src/main/java/uk/gov/register/resources/VerifiableLogResource.java
@@ -28,7 +28,7 @@ public class VerifiableLogResource {
     }
 
     @GET
-    @Path("/entry/{entry-number}/{total-entries}/merkle:sha-256")
+    @Path("/entries/{entry-number}/{total-entries}/merkle:sha-256")
     @Produces(MediaType.APPLICATION_JSON)
     @Timed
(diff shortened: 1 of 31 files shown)


gidsg:
corresponding conformance changes are in: https://github.com/openregister/conformance-test/pull/28

src/main/java/uk/gov/register/resources/VerifiableLogResource.java
```diff
@@ -29,6 +34,15 @@ public RegisterProof registerProof() {
 
     @GET
     @Path("/entry/{entry-number}/{total-entries}/merkle:sha-256")
+    public Response getProofRedirect(
```
gidsg: we need to do the redirect in this class as it owns the top level `/proof` paths.

arnau:
The 404 in HTML instead of json happens in production. Tried:

```
curl -i -H 'Accept: application/json' https://country.register.gov.uk/record/XX
```

PR #474, opened 2018-05-30
Consistent error handling

michaelabenyohai:
### Context
There is a lot of change here but much of it is renaming exceptions.

This change stemmed from wanting to improve our logging when loading RSF fails. Because there are many ways in which loading RSF can fail, this involved improving the way we catch and handle exceptions as well as where and how we log those exceptions. See the below principles I have attempted to follow (as much as possible but probably not everywhere), hopefully this will help review. 

- Rename exceptions to have more appropriate names and add descriptions to all exception classes
- Throw specific exceptions
- Catch expected/known lower-level exceptions and wrap them in a higher-level exception
- Always wrap the original exception (not just the message)
- Log the whole exception at the top, including the whole chain of exceptions
- Only use exception mappers for the high-level exceptions (which encompass low level exceptions), rather than having an exception mapper for every single exception we could possible throw
- Be explicit where a method throws an exception by adding it to the method declaration

### Guidance to review
Use the `/load-rsf` endpoint to cause exceptions (orphan entries etc.). Check output error messages as well log messages to check that the errors appear in both places.


src/main/java/uk/gov/register/exceptions/NoSuchRegisterException.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.register.exceptions;
+
+import uk.gov.register.core.RegisterId;
+
+/**
+ * Used if the definition of a particular register is missing from the register
+ */
+public class NoSuchRegisterException extends RuntimeException {
+    public NoSuchRegisterException(RegisterId registerId) {
+        super("Definition for " + registerId.value() + " register is missing");
```
gidsg: could we use `String.format` here?

src/main/java/uk/gov/register/resources/DataUpload.java
```diff
@@ -55,12 +56,13 @@ public DataUpload(ObjectReconstructor objectReconstructor, RegisterSerialisation
     @Path("/load")
     @Timed
     public void load(String payload) {
+        // TODO remove this endpoint
```
gidsg: I think this endpoint is described in the spec: https://openregister.github.io/specification/#minting so if we remove it we need to update the spec to correspond (CC @arnau) maybe best to remove this TODO for the time being?

src/main/java/uk/gov/register/providers/LoadExceptionMapper.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.register.providers;
+
+import uk.gov.register.exceptions.LoadException;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+// TODO: Remove this when we remove the /load endoint. This only exists to make tests pass.
```
gidsg: as per previous comment re: spec.

PR #475, opened 2018-06-06
fix encoding of /records/ key link

gidsg:
### Context
Previously the HTML view did not correctly handle linking where the key contains a literal backslash. e.g. https://industrial-classification-2007.cloudapps.digital/records/79.90%2F9
https://trello.com/c/B43trWa3/2025-forward-slashes-arent-url-encoded-in-api-explorer-record-id-links

fix encoding of key link where key contains literal backslash. Although Thymeleaf does URL escape `/` is excluded as it is used to form URL paths, so instead we build the encoded key and pass it into the template.

### Guidance to review
Key containing literal slash e.g. `79.90/9` should link correctly in HTML view.

### DIFF
--- a/src/main/java/uk/gov/register/core/UriTemplateLinkResolver.java
+++ b/src/main/java/uk/gov/register/core/UriTemplateLinkResolver.java
@@ -1,7 +1,10 @@
 package uk.gov.register.core;
 
 import javax.ws.rs.core.UriBuilder;
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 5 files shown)


arnau:
@gidsg keys should not allow anything out of alphanumeric, -, _. What registers are using chars out of this?

I know it's not well specified, on the to do list for improving the spec.

arnau:
See https://github.com/openregister/registers-rfcs/issues/18 for reference

PR #476, opened 2018-06-15
Add alpha phase into header

dankmitchell:
### Context
We need to tell our users that GOV.UK Registers as a service is in Aplha

- Add `Alpha` phase tag to header
- Remove the register name
- Remove navigation

### Guidance to review
View any html page

### DIFF
--- a/src/main/resources/assets/sass/_header.scss
+++ b/src/main/resources/assets/sass/_header.scss
@@ -44,7 +44,7 @@ $active-nav-color: #1d8feb;
   @extend %contain-floats;
 
   @include screen {
-    background: $black;
+    background-color: $govuk-blue;
     color: $white;
     border-bottom: 10px solid $white;
   }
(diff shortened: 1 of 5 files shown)


michaelabenyohai:
This means that unless we are on the front page there's no way of knowing what register you're looking at unless you look at the URL. Are we ok with that?

dankmitchell:
It's still on the page i.e. https://country.register.gov.uk/records

michaelabenyohai:
@dankmitchell that only exists for non-beta registers e.g. https://school-eng.alpha.openregister.org/records doesn't have it

dankmitchell:
Ah okay. Personally, I'm fine with this since its only the older `beta` registers that users know have a HTML view. We wouldn't be sending users to the API inspector and it's not part of the journey, especially the non-homepage view. Deffo a question for @conordelahunty

PR #484, opened 2018-07-17
Add ADR008 to define how to implement the items resource

arnau:
### Context

The specification defines a resource (Item collection)[https://openregister.github.io/specification/#items-resource] with an endpoint `/items`.

This ADR proposes a way to address the lack of order the set of items have by providing a first-in index.

### DIFF
--- a/doc/arch/adr-008-items-resource.md
+++ b/doc/arch/adr-008-items-resource.md
@@ -1,0 +1,81 @@
+# Decision record: Items resource implementation
+
+## Context
+
+The [Items resource](https://openregister.github.io/specification/#items-resource)
+imposes an implementation challenge regarding how to order them in a
+pagination scenario. This ADR defines a way to create a deterministic order
+for this implementation.
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


doc/arch/adr-008-items-resource.md
```diff
@@ -0,0 +1,81 @@
...
+```http
+GET /items HTTP/1.1
+Host: country.register.gov.uk
+Accept: application/json
+```
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+Link: <?cursor=3>; rel="next"
```
MatMoore: Should the API also support a link to the previous page?

arnau: Good question, my understanding of the potential usage for this endpoint is that it doesn't require a previous link. So it would be a sort-of iterator that goes forward. If we find evidence that we need the previous page, we can reconsider.

MatMoore: Ok cool. I agree it's better to leave out unless we have a strong use case.

PR #485, opened 2018-07-18
Warn about deprecated formats

MatMoore:
### Context
We're deprecating representations that are used less frequently, leaving just JSON and CSV.

The API itself should inform users of the deprecations.

Add the following headers:

- `Link: <https://....json>; rel="alternate"` for .yaml or .ttl
- `Link: <https://....csv>; rel="alternate"` for any .tsv or .xlsx
- `Warning: 299 - "Miscellaneous Persistent Warning" "yaml is deprecated
  and will be removed"` for all deprecated formats

Trello: https://trello.com/c/DyKxUdyA/2638-add-format-change-comms-to-api

See also:
- https://tools.ietf.org/html/rfc7234#section-5.5.7
- https://www.iana.org/assignments/link-relations/link-relations.xhtml

### Guidance to review
We have to use `httpServletResponse` instead of the usual `ContainerResponseContext`, because the pagination code is already adding `Link` headers this way, and when Jersey writes the response, it chooses to overwrite any headers that are already set on `httpServletResponse` with the same name as those added to the `ContainerResponseContext`.

Questions for reviewers:

- Do the warning messages make sense?
- Is the logic for generating an alternate URL correct?
- In particular, is the `type` the best parameter name to use for the media type?
- How can I make my java more idiomatic?

### DIFF
--- a/src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
+++ b/src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
@@ -77,7 +77,7 @@ public class DeprecatedFormatFilter implements ContainerResponseFilter {
     }
 
     private String warningHeader(String deprecatedType) {
-        return "299 - \"Miscellaneous Persistent Warning\" \"" + deprecatedType + " is deprecated and will be removed. See \"Link\" header for a format to use instead.\"";
+        return "299 - \"" + deprecatedType + " is deprecated and will be removed. See 'Link' header for a format to use instead.\"";
     }
 
     private boolean recommendJson(MediaType mediaType) {
--- a/src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
+++ b/src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
@@ -42,10 +42,10 @@ public class DeprecatedFormatFilter implements ContainerResponseFilter {
 
         if (recommendJson(mediaType)) {
             alternateUri = alternateLink(requestedUri, ".json");
-            alternateType = MediaType.APPLICATION_JSON;
+            alternateType = "application/json";
         } else if (recommendCsv(mediaType)) {
             alternateUri = alternateLink(requestedUri, ".csv");
-            alternateType = ExtraMediaType.TEXT_CSV;
+            alternateType = "text/csv";
         } else {
             return;
         }
--- a/src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
+++ b/src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
@@ -1,21 +1,30 @@
 package uk.gov.register.filters;
 
+import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerResponseContext;
 import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriInfo;
 import javax.ws.rs.ext.Provider;
 import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 
 import com.google.common.net.HttpHeaders;
+import uk.gov.register.resources.HttpServletResponseAdapter;
 import uk.gov.register.views.representations.ExtraMediaType;
 
 @Provider
 public class DeprecatedFormatFilter implements ContainerResponseFilter {
+
+    @Context
+    HttpServletResponse httpServletResponse;
+
     @Override
     public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
         final UriInfo uriInfo = requestContext.getUriInfo();
@@ -39,10 +48,13 @@ public class DeprecatedFormatFilter implements ContainerResponseFilter {
             return;
         }
 
-        headers.add(
-                HttpHeaders.LINK,
-                linkHeader(alternateType, alternateUri)
-        );
+        HttpServletResponseAdapter httpServletResponseAdapter = new HttpServletResponseAdapter(httpServletResponse);
+
+        Map<String, String> extra = new HashMap<>();
+        extra.put("rel", "alternate");
+        extra.put("type", alternateType);
+        httpServletResponseAdapter.addLinkHeader(extra, alternateUri);
+
         headers.add(
                 HttpHeaders.WARNING,
                 warningHeader(mediaType.getSubtype())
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
```diff
@@ -0,0 +1,75 @@
...
+        final String newPath = path.replaceFirst("(/)?$", recommendedExtension + "$1");
+
+        builder.replacePath(newPath);
+
+        return builder.toTemplate();
+    }
+
+    private void addWarning(MultivaluedMap<String, Object> headers, String deprecatedType, String alternate) {
+        headers.add(HttpHeaders.LINK, alternate + "; rel=\"alternate\"");
+        headers.add(HttpHeaders.WARNING, "299 - \"Miscellaneous Persistent Warning\" \"" + deprecatedType + " is deprecated and will be removed\"");
```
arnau: Perhaps we could extend this message suggesting they check the alternate header for a suitable alternative?

MatMoore:
Adding the content type makes sense, I'll add that and extend the warning message.

I'm not sure whether adding additional alternates is worth it, particularly if we're providing multiple URLs for the same resource.

I was also wondering whether we should have alternate links on every success response, not just the deprecated formats. Each representation could link to each (non-deprecated) representation, instead of each deprecated representation linking to a preferred alternative.

It would help discoverability more generally, but it would make the deprecation notice less clear, so not sure if it's worth the added complexity. If we're implementing like this as a filter, that would mean there's two places we have to define (or inspect?) the supported content types for every endpoint.

arnau:
Ok, let's keep it simple and offer a single alternative with type.

After the deprecation period is over and we remove the warnings, we can consider adding alternate links for all resources.

MatMoore:
I think this is ready for review again now.

I've changed the format of the warning header, as I'd inadvertently put more than one `warning-text` value in.

Example response:

```
curl -I 'localhost:8080/records.yaml'
HTTP/1.1 200 OK
Date: Fri, 20 Jul 2018 11:21:00 GMT
Link: <?page-index=2&page-size=100>; rel="next",<http://localhost:8080/records.json>; rel="alternate"; type="application/json"
Warning: 299 - "yaml is deprecated and will be removed. See 'Link' header for a format to use instead."
Content-Type: text/yaml;charset=UTF-8
Content-Security-Policy: default-src 'self' www.google-analytics.com; connect-src 'self' www.google-analytics.com; img-src 'self' www.google-analytics.com; script-src 'self' www.google-analytics.com; font-src data:;
X-Content-Type-Options: nosniff
Strict-Transport-Security: max-age=10886400; includeSubDomains
X-XSS-Protection: 1; mode=block
Cache-Control: no-transform
Content-Length: 0
```

I noticed that the `alternate` header is absolute whereas the `rel` header is relative but I'm not sure if this matters?

I'm also unsure about the correctness of the link headers with a type parameter. @arnau I followed your example of `; type="application/json` but I couldn't find any examples elsewhere to verify that "type" is the correct parameter name here. All I found is [RFC 2068](http://freesoft.org/CIE/RFC/2068/245.htm) which allows for arbitrary key value pairs.

arnau:
re. relative vs absolute: I tens to prefer absolute links in headers but historically registers has used relative links. In the case of the alternate, because it's a different url not a change of parameter you should at least have an absolute path (i.e. `/records/foo.yaml`) instead of a relative one.

re. "type", this attribute name is the one used by HTML when using `rel="alternate"` and as far as I can tell W3C tends to use it this way but I haven't found any source that makes it formal.

src/main/java/uk/gov/register/resources/HttpServletResponseAdapter.java
```diff
@@ -5,25 +5,44 @@
...
 
-    HttpServletResponseAdapter(HttpServletResponse httpServletResponse) {
+    public HttpServletResponseAdapter(HttpServletResponse httpServletResponse) {
         this.httpServletResponse = httpServletResponse;
     }
 
-    void addInlineContentDispositionHeader(String fileName) {
+    public void addInlineContentDispositionHeader(String fileName) {
         ContentDisposition contentDisposition = ContentDisposition.type("inline").fileName(fileName).build();
         httpServletResponse.addHeader(HttpHeaders.CONTENT_DISPOSITION, contentDisposition.toString());
```
michaelabenyohai: We could convert this to use `setHeader`

src/main/java/uk/gov/register/resources/HttpServletResponseAdapter.java
```diff
@@ -5,25 +5,44 @@
...
     }
 
-    void addLinkHeader(String rel, String value) {
+    public void addLinkHeader(String rel, String value) {
```
michaelabenyohai: Wonder whether it's worth renaming all the methods here to be `set...` rather than `add...` to be consistent and not mislead?

src/main/java/uk/gov/register/filters/DeprecatedFormatFilter.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private boolean recommendCsv(MediaType mediaType) {
+        return mediaType.equals(ExtraMediaType.TEXT_TSV_TYPE) ||
+                mediaType.equals(ExtraMediaType.APPLICATION_SPREADSHEET_TYPE);
+    }
+
+    private Optional<MediaType> extractMediaType(MultivaluedMap<String, Object> headers) {
+        try {
+            MediaType mediaType = (MediaType) headers.getFirst(HttpHeaders.CONTENT_TYPE);
```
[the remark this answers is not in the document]
MatMoore: I'm not sure about this actually... it feels safer to use the MediaType object for the comparison as I'm not sure if the string representation of the value is always going to be exactly what I expected. If I use MediaType.equals it should do the right thing even if there's differences in case or whitespace.

Let me know if there's a better way to accomplish this though.

MatMoore:
@michaelabenyohai, I've implemented all the suggestions one (see comment above). Can you take another look?

MatMoore:
I think there might be a small bug actually:

The CSV values are generated as:

`<?page-index=2&page-size=100>; rel="next",<http://localhost:8080/records.csv>; rel="alternate"; type="text/csv; charset=UTF-8"`

(note the quoting of the type part)

I think that should be: 

`<?page-index=2&page-size=100>; rel="next",<http://localhost:8080/records.csv>; rel="alternate"; type="text/csv"; charset="UTF-8"`

or possibly 

`<?page-index=2&page-size=100>; rel="next",<http://localhost:8080/records.csv>; rel="alternate"; type=text/csv; charset=UTF-8`

MatMoore:
I've pushed one more commit to deal with the above problem.

arnau:
@MatMoore good catch, I think it should be:

```
<?page-index=2&page-size=100>; rel="next",<http://localhost:8080/records.csv>; rel="alternate"; type="text/csv"; charset="UTF-8"
```
As it keeps consistency in quoting values like we do with `rel`.

MatMoore:
@arnau I've solved this by just stripping out the charset part. I don't think we need it here?

arnau:
I did a quick check to https://www.w3.org/TR/tabular-data-model/ and I think you are right as UTF-8 is the default charset there.

PR #490, opened 2018-08-03
fix dependency check

gidsg:
### Context
dependency check was failing

- Surpress erroneous stax-api warning
- Remove tomcat suppressions which no longer cause failure

### Guidance to review
`./gradlew dependencyCheckAnalyze` should be successful.


gidsg:
It appears to be erroneous as the actual CVE is about an node module: https://nvd.nist.gov/vuln/detail/CVE-2017-16224

PR #491, opened 2018-08-03
Correct version history link header

MatMoore:
### Context
The `Link` header for `/records/{key}` with relation
`rel="version-history"` uses the old url `/record/{key}/entries` instead
of the new one `/records/{key}/entries`.

Trello: https://trello.com/c/333bUFTY/2733-the-link-relation-version-history-shows-the-wrong-old-url

Add an 's'

### Guidance to review
The link should point to the correct entries resource.


MatMoore:
Yeah I need to rebase it.

PR #493, opened 2018-08-14
switch to LocalDate for date validation parser

gidsg:
### Context
Previous date validator allowed values which should be illegal in the [spec](https://openregister.github.io/specification/#datetime-datatype) e.g. `41365` and `2018-1`

Switch to Java 8 [LocalDate](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html) for date validation. 
Switch to [parseDateStrictly](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/time/DateUtils.html#parseDateStrictly(java.lang.String,%20java.lang.String[])) for ordering validation.

Note, the reason this PR switches to use a list of formatters rather than parsing straight from the format string is that `LocalDate` always requires a year, month and date to be specified so we use the formatter to default them when they are not included in our permitted pattern.

### Guidance to review
Tests should pass in Travis CI


MatMoore:
LGTM but I'm unfamiliar with the Java 8 API so I'll defer to @michaelabenyohai or @arnau to approve/reject.

PR #495, opened 2018-08-20
ADR009: Trailing slash in HTTP resources

arnau:
### Context

Currently, HTTP resources have many URLs representing the same thing. This is
not an ideal scenario for search engine indexing or caching.

It proposes a canonical URL per resource.

### Guidance to review

It should cover all cases.

### DIFF
--- a/doc/arch/adr-009-trailing-slash.md
+++ b/doc/arch/adr-009-trailing-slash.md
@@ -20,6 +20,9 @@ This ADR proposes a canonical URL and a way to consolidate around it.
 The canonical path for a resource MUST NOT have a trailing slash. Aliases MUST
 redirect (301) to the canonical path.
 
+Also, the `Link` header for aliases should add the canonical URL using the
+`rel="canonical"` ([RFC6596](https://tools.ietf.org/html/rfc6596)).
+
 For example, the `GB` record has the canonical path:
 
 ```
--- a/doc/arch/adr-009-trailing-slash.md
+++ b/doc/arch/adr-009-trailing-slash.md
@@ -1,0 +1,73 @@
+# Decision record: Trailing slash in HTTP resources
+
+## Context
+
+Currently, HTTP resources have many URLs representing the same thing. This is
+not an ideal scenario for search engine indexing or caching. For example, the
+record `GB` resource can be accessed via:
+
+* `GET /records/GB`
+* `GET /records/GB/`
+* `GET /records/GB.json`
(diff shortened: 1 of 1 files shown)


gidsg:
We could also look at defining `link rel="canonical"` for example to catch cases where non-supported query parameters are used, we would want to point search engines to a single valid URL. Perhaps this would be a separate ADR or RFC though?

doc/arch/adr-009-trailing-slash.md
```diff
@@ -0,0 +1,73 @@
...
+
+Aliases like `/records/GB/` and `/records/GB.json/` MUST redirect to their
+respective canonical resources.
+
+### Paginated resources
+
+The first page for the records list has the canonical path:
+
+```
+/records?page-index=1&page-size=100
```
gidsg: So this means `/records` would 301 to `/records?page-index=1&page-size=100`? I guess this would close off for example the possibility to change the default page size in future as clients would remember this?

arnau: What do you mean by "would remember this"?

arnau: > We could also look at defining link rel="canonical" for example to catch cases where non-supported query parameters are used, we would want to point search engines to a single valid URL. Perhaps this would be a separate ADR or RFC though?

This could be part of this ADR and probably a non-normative note in the specification.

arnau:
> We could also look at defining link rel="canonical" for example to catch cases where non-supported query parameters are used, we would want to point search engines to a single valid URL. Perhaps this would be a separate ADR or RFC though?

Probably belongs to this ADR and a non-normative note in the specification.

doc/arch/adr-009-trailing-slash.md
gidsg: `301` implies the client does not have to check back with the origin, v.s. `302` where it will check every time.

gidsg: I am not sure about this for index pages, it is quite nice to have a canonical index page without defaults specified. Do you have any other examples of popular APIs that redirect paginated routes to the first page URL like this?
If not, I would rather get the trailing slash part approved but not the paginated resources part.

arnau: I don't have a popular API example, I'll try to find one. I don't see the "quite nice" part to be honest but I think you have a fair point on focusing on trailing slashes for now and discuss further about pagination.

PR #499, opened 2018-10-05
Update to Java 9 [DO NOT MERGE]

gidsg:
### Context
We need to get off Java 8 before it goes out of support. Unfortunately we are blocked on Java 10+ by upstream issues e.g. https://github.com/dropwizard/dropwizard/issues/2450 so this PR moves to Java 9 which is out of support (hence do not merge) but gets us closer to the eventual goal of Java 11 LTS

This involved updating many dependencies to versions that support Java 9, changes include:
* Fix deprecation warnings
* Update to Dropwizard 1.3.6
* Update to Gradle 4.10
* Add dependencies on Java modules that were removed from core as part of modularisation process
* fork compass-gradle-plugin to publish version which supports Gradle 4.10 (see https://github.com/robfletcher/gradle-compass/issues/77)
* Update Travis CI and CloudFoundry java version

### Guidance to review
Tests should pass on Travis CI, app should run locally if you are using Java 9 (I suggest using `jenv` to allow for multiple system JDK versions)


gidsg:
closing as we don't intend to merge this. Will review again when upstream issue resolution may allow us to get to Java 11.

PR #502, opened 2018-10-10
make JSON default representation if no format specified

gidsg:
### Context
https://trello.com/c/m8sbtMLi/2956-make-json-the-default-format-when-accept-and-extension-are-missing

Make JSON default format when no preference is expressed. The mechanism to do this is simply to declare JSON resources higher up than HTML ones which then makes them default (if no media type is specified.)

### Guidance to review
Tests should pass in Travis CI
`curl http://localhost:8080/records` should return JSON


gidsg:
The tests are flaky on travis CI in a way I cannot reproduce locally, so going to close this.

PR #503, opened 2018-10-15
Implements ADR009: Trailing slash in HTTP resources 

gidsg:
### Context
https://github.com/openregister/openregister-java/blob/master/doc/arch/adr-009-trailing-slash.md

Redirect trailing slash paths to non-trailing slash

### Guidance to review
Note this does not implement `rel=canonical` as suggested in the ADR, as while they were in the same ADR I would rather implement `rel=canonical` in a separate PR.

### DIFF
--- a/src/main/java/uk/gov/register/filters/StripTrailingSlashRedirectFilter.java
+++ b/src/main/java/uk/gov/register/filters/StripTrailingSlashRedirectFilter.java
@@ -1,0 +1,52 @@
+package uk.gov.register.filters;
+
+import com.google.common.net.HttpHeaders;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class StripTrailingSlashRedirectFilter implements javax.servlet.Filter {
+
+    private static String removeLastChar(String str) {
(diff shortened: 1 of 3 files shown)


arnau:
> Note this does not implement rel=canonical as suggested in the ADR, as while they were in the same ADR I would rather implement rel=canonical in a separate PR.

👍

src/main/java/uk/gov/register/filters/StripTrailingSlashRedirectFilter.java
```diff
@@ -0,0 +1,52 @@
...
+    public void destroy() { }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest httpRequest = (HttpServletRequest)request;
+
+        String path = httpRequest.getRequestURI();
+        if (path.endsWith("/") && path.length() > 1) {
+        int serverPort = httpRequest.getServerPort();
+            StringBuilder location = new StringBuilder();
```
arnau: I would've thought there were a standard URL constructor/builder so we wouldn't have to have to construct it ourselves :(

gidsg: good point, I have switched to use a URIBuilder now.

PR #505, opened 2018-10-15
[WIP][SPIKE] Replace items with blobs

MatMoore:
### Context
- We previously decided to replace "items" with "blobs" in the data model and REST API (https://github.com/openregister/registers-rfcs/pull/29)
- We are currently working out how to version the API (https://trello.com/c/Rmy3TVcE/2957-define-versioning-strategy)

This is a throwaway spike to see how we could do this in practice.

- Deprecate /items/*
- Introduce /v2/blobs/*
- Update the HTML templates
- Rename classes internally (only partially done)

### Guidance to review
- No need for a detailed review just yet
- Each resource currently has its own version: does this make sense or should we version the whole API together?
- `DataDownloadFunctionalTest` seems to be flaky and I don't understand why (I think this happened the last time I worked on ORJ as well). 2 tests against root hashes fail on this branch but pass when run individually.

I've deliberately not changed the database schemas, since we are likely to change this anyway when removing indexes. I've also not changed the RSF command as I think RSF should be evolved separately (we can raise a separate RFC for this)


MatMoore:
Ending this spike here. Write up: https://docs.google.com/document/d/1GXAIsZ96eF4HzhIA9rKvZzKT-XwZ4DXA_gtWeusG4Yo/edit#heading=h.mffdf92r2io0

PR #509, opened 2018-10-23
@Ignore flaky tests

gidsg:
### Context
These tests behave inconsistently between local runs and CI and were creating false positive concerns.

`@Ignore` these tests

### Guidance to review
Tests should pass in CI. We may wish to make a backlog card to review if we can either make them reliable or delete them entirely.


arnau:
👍 to a trello card so we make sure we have a plan to solve it

gidsg:
OK I raised a backlog card: https://trello.com/c/js515HVz/1471-review-flaky-tests

PR #511, opened 2018-10-25
[WIP] Pre index removal manual migrations

MatMoore:
### Context
Master is not currently deployable, due to a problem with long running migrations.

At the moment we automatically run migrations when we start the app. This causes problems when migrations take a long time to run, because when doing a zero downtime deployment, cloud foundry will kill the app if it doesn't respond to requests within a certain time.

The migration on master is backwards compatible with the old code. The new code (to remove indexes) requires the migration to run.

Do not run the migrations on app startup. Run them 

### Guidance to review
The alternative to this approach is to do a regular `cf push` deployment, with downtime.

However I thought it was worth creating a separate command to manually run migrations, since this is something that is likely to happen again. However, this actually creates some logistical issues, because we need to remove the current code that runs migrations as part of the serve command, and trigger this task elsewhere in the build, before the app is run. I think this needs more thought, so I don't propose merging this branch.

In order to deploy master now, I'd like to deploy this branch as a hotfix with automatic migrations disabled, run the migration manually, and then deploy master as normal.

The code on this branch is the same as the version currently deployed, with indexes still present.

The functional tests apparently rely on migrations being run at app startup to work. It will take me forever to work out how to fix this, so I've just commented out the failing ones for the purposes of running this migration.

### DIFF
--- a/src/main/java/uk/gov/migration/V__13_Add_item_to_entry_table.java
+++ b/src/main/java/uk/gov/migration/V__13_Add_item_to_entry_table.java
@@ -1,0 +1,39 @@
+package uk.gov.migration;
+
+import org.flywaydb.core.api.migration.MigrationChecksumProvider;
+import org.flywaydb.core.api.migration.jdbc.BaseJdbcMigration;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.Statement;
+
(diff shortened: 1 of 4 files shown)


MatMoore:
These flaky tests are making this take forever, so I'm going to abandon this approach and try changing the healthcheck instead.

PR #515, opened 2018-10-29
Redirect /v1/ to the unversioned URLs

MatMoore:
### Context
We are going to start versioning URLs in line with the [GDS API standards](https://www.gov.uk/guidance/gds-api-technical-and-data-standards#when-you-need-to-make-a-backwards-incompatible-change).

Since we didn't have `/v1/` in the URLs from the beginning, I think it makes sense to keep all the existing resources as they are, but redirect `/v1/` URLs. That way any users who try to construct the v1 URLs based on newer documentation will still get sent to the unversioned resources.

I used a temporary redirect rather than a permanent one because we're redirecting to something with a fixed life expectancy.

### Guidance to review
- Does this make sense?
- Is there a cleaner way to do this?
- I'm working on the assumption that there will be a v2 of every resource, even if nothing has changed from v1.
- After this my plan is to move the resource classes into a package per version (and probably the same with the views)

### DIFF
--- a/src/main/java/uk/gov/register/resources/RedirectResource.java
+++ b/src/main/java/uk/gov/register/resources/RedirectResource.java
@@ -1,8 +1,15 @@
 package uk.gov.register.resources;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
+import javax.ws.rs.HEAD;
+import javax.ws.rs.HttpMethod;
+import javax.ws.rs.OPTIONS;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/resources/RedirectResource.java
```diff
@@ -59,12 +108,17 @@ public Response getRecordEntriesRedirect(
         return redirectByPath(request, "/record/", "/records/");
     }
 
-
-
     private static ResponseBuilder movedPermanently(URI location) {
         return status(Response.Status.MOVED_PERMANENTLY).location(location);
     }
 
+    public static Response temporaryRedirectByPath(@Context HttpServletRequest request, String oldPath, String newPath) {
```
gidsg: we could use overloading to DRY this up a bit. i.e. add a version of `redirectByPath` that takes a `responseStatus` argument and default the current one to `MOVED_PERMANENTLY`.

PR #517, opened 2018-10-31
Add new v2 endpoints

MatMoore:
### Context
This builds on top of https://github.com/openregister/openregister-java/pull/515 so I'll need to rebase this once that's merged.

Introduce a /dev/ prefix so we can serve multiple versions of the API at the same time.

Anyone using these endpoints should expect functionality to be
added/removed/changed. Once this version is stable, we will rename
it to /v2/.

As part of this change I've moved the main resource classes (i.e. stuff
that's part of the specified API) to subpackages by version.

I've made IndexSizePagination public so it can be used in both of these
packages (previously it was package private).

As a follow up to this work, we will need to do something with the
conformance tests so that we have the ability to test multiple
versions of the API against their corresponding specification.

### Guidance to review
- The unversioned (v1) endpoints should still work
- You should be able to put /dev/ in front of any API URL
- I decided to make v1 inherit from v2 rather than the other way round because I think v2 is the more interesting implementation from the perspective of someone auditing the code (as it will be the most up to date) so I want this to be easy to read without having to look at legacy code as well.

### DIFF
--- a/src/main/java/uk/gov/register/resources/v2/EntryResource.java
+++ b/src/main/java/uk/gov/register/resources/v2/EntryResource.java
@@ -1,4 +1,4 @@
-package uk.gov.register.resources;
+package uk.gov.register.resources.v2;
 
 import com.codahale.metrics.annotation.Timed;
 import io.dropwizard.jersey.params.IntParam;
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/register/resources/v2/EntryResource.java
```diff
@@ -34,7 +37,7 @@ public EntryResource(RegisterReadOnly register, ViewFactory viewFactory, Request
         this.register = register;
         this.viewFactory = viewFactory;
         this.requestContext = requestContext;
-        this.httpServletResponseAdapter = new HttpServletResponseAdapter(requestContext.httpServletResponse);
+        this.httpServletResponseAdapter = new HttpServletResponseAdapter(requestContext.getHttpServletResponse());
```
gidsg: what's the purpose of this change? Was it intentional?

MatMoore: It's because ReqeustContext is now in a different package to the v1 and v2 packages, so the field can't be accessed directly anymore.

PR #518, opened 2018-11-02
Add blob resource

MatMoore:
Trello: https://trello.com/c/gRvrLxdY/2987-implement-rfc0016-blob-resource

### Context
This PR implements RFC 16 - Blob Resource (https://github.com/openregister/registers-rfcs/pull/29)

It also adds the list endpoint (GET /blobs), which we had never implemented. This was previously reported in https://github.com/openregister/specification/issues/9. The consequence of not having /blobs is that if you are using the blobs and entries part of the API, you need to make a request for every item you retrieve, rather than fetching multiple items in one go.

I've currently restricted this to the first 100 items, with the goal of implementing pagination in a later sprint. Since there are other issues with pagination it will be easier for me to work on it all at the same time.

- [x] Rename /items -> /blobs
- [x] Add missing GET /blobs endpoint (this only supports CSV, JSON, HTML but it is backported to V1 as well)
- [x] Rename item-hash -> blob-hash in entries endpoints
- [x] Rename item-hash -> blob-hash in /record/entries
- [x] <strike>Rename the internal directory “item” inside the archive (/download-register) to “blob”.</strike> This is not needed in v2 as we are [removing the endpoint](https://github.com/openregister/registers-rfcs/blob/master/content/archive-resource/index.md).
- [x] Resource unit tests

### Follow up work I'll raise another PR for or add to the backlog
- [ ] Conformance tests
- [ ] Tech docs
- [ ] Implementing the HTML views for V2 endpoints such that links remember the version you are on
- [ ] Pagination for /blobs
- [ ] Updating the link headers in responses to be version-aware

### Guidance to review
- This should be reviewed one commit at a time. If it's easier I can go through it in person.
- There is a separate resource for each version of a resource, and there are separate view classes for every entity where the representations have changed between versions.


src/main/java/uk/gov/register/resources/v2/BlobResource.java
```diff
@@ -47,13 +47,13 @@ public ItemResource(RegisterReadOnly register, ViewFactory viewFactory) {
             ExtraMediaType.APPLICATION_SPREADSHEET
     })
     @Timed
-    public ItemView getItemDataByHex(@PathParam("item-hash") String itemHash) throws FieldConversionException {
-        return getItem(itemHash).map(viewFactory::getItemMediaView)
-                .orElseThrow(() -> new NotFoundException("No item found with item hash: " + itemHash));
+    public ItemView getBlobDataByHex(@PathParam("blob-hash") String blobHash) throws FieldConversionException {
+        return getBlob(blobHash).map(viewFactory::getItemMediaView)
+                .orElseThrow(() -> new NotFoundException("No item found with item hash: " + blobHash));
```
gidsg: should this text be _No item found with blob hash:_?

MatMoore: Yes it should - I think there are a few of these messages that need changing

MatMoore:
@gidsg there is at least one bug I still need to fix, but I think this is in a reviewable state. Can you take a look tomorrow?

MatMoore:
I've taken out the web view part of this PR, so it's now ready to review.

src/main/java/uk/gov/register/db/ItemQueryDAO.java
```diff
@@ -21,11 +21,14 @@
     @RegisterMapper(ItemMapper.class)
     Optional<Item> getItemBySHA256(@Bind("sha256hex") String sha256Hash, @Define("schema") String schema );
 
-    //Note: This is fine for small data registers like country
     @SqlQuery("select sha256hex, content from \"<schema>\".item order by sha256hex")
     @RegisterMapper(ItemMapper.class)
     Collection<Item> getAllItemsNoPagination( @Define("schema") String schema );
 
+    @SqlQuery("select sha256hex, content from \"<schema>\".item i where exists(select 1 from \"<schema>\".<entry_table> e where i.sha256hex = e.sha256hex) order by sha256hex")
```
gidsg: given that the `sha256hex` is not a human meaningful string is there any reason to sort by it? What would an alternative sort option be?

src/test/java/uk/gov/register/resources/v1/EntryResourceTest.java
```diff
@@ -17,15 +17,6 @@
 
 @RunWith(MockitoJUnitRunner.class)
 public class EntryResourceTest {
-
-    @Test
-    public void findByEntryNumberHtmlSupportsHtml() throws Exception {
```
gidsg: I was not aware we removed this functionality for V1, so should the test still work?

src/test/java/uk/gov/register/views/v2/EntryListViewTest.java
```diff
@@ -0,0 +1,82 @@
...
+            EntryType.user
+    );
+    private final EntryListView view = new EntryListView(ImmutableList.of(entry1, entry2));
+
+    @Test
+    public void jsonRepresentation() throws JsonProcessingException {
+        String result = objectMapper.writeValueAsString(view);
+
+        assertThat(result, equalTo("[" +
+                "{" +
```
gidsg: might be more stable and readable to make assertions on the JSON value using something like Jackson rather than a string blob? Do not feel very strongly about this though.

gidsg: ...apparently in Java 12 string literals are coming which would make these kinda things much nicer: https://openjdk.java.net/jeps/326

src/main/resources/templates/fragments/v2-entry-table.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
```
gidsg: I don't think this file is actually used now you have removed the HTML view from V2?

src/main/java/uk/gov/register/resources/v1/ItemResource.java
```diff
@@ -1,15 +1,52 @@
...
 import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.NotFoundException;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import java.util.Collection;
+import java.util.stream.Collectors;
 
 @Path("/items")
```
gidsg: I think it may be prudent to remove or comment out this route until it is feature complete to include in V1.

src/main/resources/templates/fragments/v2-entry-table.html
MatMoore: Yes, this should be removed

src/main/java/uk/gov/register/resources/v1/ItemResource.java
MatMoore: Will do

MatMoore:
While looking at the conformance tests I've noticed a bug in /records/foo/entries so I will need to fix this before merging

src/test/java/uk/gov/register/resources/v1/EntryResourceTest.java
MatMoore: I must have removed this by accident, I'll add it back

src/main/java/uk/gov/register/db/ItemQueryDAO.java
MatMoore: The only reason to use it is that we need some kind of ordering of the blobs in order to split them over pages.

Although I agree that the hash doesn't make a lot of sense. If blobs get added to the register while you are paging through it you will get strange results.

Actually I just noticed today that there is an ADR for this here: https://github.com/openregister/openregister-java/blob/master/doc/arch/adr-008-items-resource.md

This suggests ordering items by insertion order, which would mean adding a serial field to the items table, but seems like a better approach than what I've done here. Would it be ok to comment this endpoint out for now and address this in a future PR? It already needs further work to add the pagination.

gidsg: yes, that is fine, can we make a trello card for it so we don't forget?

MatMoore: Turns out there was already a card, https://trello.com/c/FiMDw8PF/689-implement-blobs-endpoint-in-orj - but it's not actually in scope for this quarter at the moment.

I would like to get it done before then if we have time though, because it makes much easier to fetch all entries and blobs for a register.

PR #519, opened 2018-11-08
Test both versions of the API

MatMoore:
### Context
This is a follow up to https://github.com/openregister/conformance-test/pull/31 to make sure the build runs tests for both API versions.

- Use the latest conformance tests
- Run tests for both api versions

### Guidance to review
`./gradlew check` should run the conformance tests for both versions of the API.

I'm not familiar with gradle so let me know if there's a better way of specifying the relationship between the dependencies here.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -169,18 +169,24 @@ task installConformanceTestDeps(type: Exec) {
     commandLine '.venv/bin/pip', 'install', '-r', 'requirements.txt'
 }
 
-task runConformanceTests(type: Exec) {
+task runConformanceTestsV1(type: Exec) {
     dependsOn << assemble
     dependsOn << startAppForConformance
     dependsOn << loadSchoolDataForConformance
     dependsOn << installConformanceTestDeps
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -169,18 +169,25 @@ task installConformanceTestDeps(type: Exec) {
...
     dependsOn << loadSchoolDataForConformance
     dependsOn << installConformanceTestDeps
+    mustRunAfter test
+    commandLine '.venv/bin/openregister-conformance', '--api-version', '1', '--no-https', '--register', 'school', 'http://localhost:9090/v1/', '--register-domain', 'test.register.gov.uk'
+}
+
+task runConformanceTestsV2(type: Exec) {
+    dependsOn << runConformanceTestsV1
     finalizedBy stopAppForConformance
     mustRunAfter test
```
gidsg: I think this line is redundant as this already depends on `runConformanceTestsV1` which `mustRunAfter test`

PR #521, opened 2018-11-13
Remove YAML internal dependence

gidsg:
### Context
Previously a register loaded configuration for Fields, Registers and Public Bodies in YAML format. The YAML data for Registers and Fields was derived from the ORJ YAML serialisation. As this is being deprecated we need to read this config in an alternative format.

Use JSON format for configuration files instead of YAML

### Guidance to review
App should work as before. Note: this should not be merged until the corresponding change is made in Ansible to write JSON to S3 instead of YAML.


gidsg:
@MatMoore I believe the only connected infrastructure is the config in S3. For alpha & beta this is updated by an Ansible task and on discovery it is updated manually. This will need to be done before this is merged.

PR #523, opened 2018-11-14
Remove spreadsheet, TSV, YAML support

MatMoore:
### Context
GOV.UK Registers will stop supporting data formats other than JSON and CSV from 1 December 2018.

- Remove annotations from all the endpoints
- Remove the resource writers for these formats
- Remove `ExtraMediaType` constants
- Remove `DeprecatedFormatFilter`
- Remove these formats from the `UriDataFormatFilter`

### Guidance to review
- Look for anything I've missed
- Other formats should still work
- It seems like we still use TSV and YAML for some unrelated things like test fixtures, which I've left as is


gidsg:
I believe `apachePoi` gradle dependency is used for XLSX and can be safely removed.

gidsg:
We need to update conformance-test dependency to a matching version.

MatMoore:
I think we can remove apache jena as well (turtle). I'll look into this tomorrow.

PR #524, opened 2018-11-15
ADR011: Propose a method to handle the root paths consistently

arnau:
### Context

As part of the work to release the next major version, there are three routes
that need special handling: `/`, `/v1` and `/v2`.

This ADR proposes a behaviour for each one of them.

### DIFF
--- a/doc/arch/adr-011-root-path-handling.md
+++ b/doc/arch/adr-011-root-path-handling.md
@@ -11,19 +11,25 @@ This ADR proposes a behaviour for each one of them.
 
 ### Root (`/`)
 
-When HTML is requested, it should redirect (301) to `https://registers.service.gov.uk/registers/{register_id}`.
+When HTML is requested, it should return the exact same HTML as per now.
 
 Otherwise it should redirect (301) to the Register resource
-(`/{version}/register`) for the current version of the API in the requested
-format; JSON if unspecified.
+(`/{version}/register`) for the requested version in JSON regardless of the
+format specified.
 
-### Version root (`/v1`, `/v2`)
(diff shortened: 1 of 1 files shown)


MatMoore:
The non HTML redirects make sense, but I don't agree with the part about HTML.

If we are dropping the HTML views in version 2, then this would mean that you can request HTML, but only via this one redirect.

Also, I don't think it's a good idea for ORJ to redirect to registers.service.gov.uk if it's not the deployment that is providing registers.service.gov.uk. It's conceivable that somebody else might want to run ORJ and not have anything to do with registers.service.gov.uk.

I know ORJ currently links to this URL in the web view for v1, but I disagree with that behaviour as well.

I suggest that instead we make v1 redirect to `/register` regardless of content type. At the moment there is no HTML view for `/register`, even in the old version, so this would display JSON to users requesting HTML.

gidsg:
@arnau I don't like the redirect to the service domain, as I think it would cause confusion. i.e. in the docs we talk about the base path and if people try hitting that in their browser and end up at the service domain I don't think it's very intuitive.

I would be OK to maintain the existing HTML view on the root path for the lifetime of V1. For /v2 and once /v1 is deprecated perhaps `307` to `/register` as this would reserve us the right to do something else with the root path in future.

arnau:
> @arnau I don't like the redirect to the service domain, as I think it would cause confusion

I disagree with the statement but given that you two agree, I'll change the proposal. This proposal must be pragmatic not seeking perfection.

arnau:
As per the `307`, I'm going to stick with 301. It is known, reliable and the intention is clear.

PR #527, opened 2018-11-16
Implement ADR 11

MatMoore:
### Context
See https://github.com/openregister/openregister-java/pull/524
https://trello.com/c/m2QMFk4c/3083-stuff-missing-from-v2-api

- Redirect the versioned base path to /register when requesting CSV/JSON
- Redirect /v1 to / when requesting HTML

### Guidance to review
- Check redirect behaviour against the ADR
- I had to change the paths in the v2 resources for it to pick up the new route

### DIFF
--- a/src/test/java/uk/gov/register/resources/RedirectResourceTest.java
+++ b/src/test/java/uk/gov/register/resources/RedirectResourceTest.java
@@ -8,6 +8,7 @@ import uk.gov.register.functional.app.RegistersRuleDisableFollowRedirects;
 import uk.gov.register.functional.app.RsfRegisterDefinition;
 import uk.gov.register.functional.app.TestRegister;
 
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import static javax.ws.rs.core.MediaType.WILDCARD;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/register/resources/RedirectResourceTest.java
```diff
@@ -40,23 +41,44 @@ public void getItemRedirect() throws Exception {
...
+        Response response = register.getRequest(TestRegister.register, "/next.json", WILDCARD);
+        assertThat(response.getStatus(), equalTo(301));
+        assertThat(response.getLocation().getPath(), equalTo("/next/register.json"));
+    }
+
     @Test
     public void getV1ItemRedirect() throws Exception {
         Response response = register.getRequest(TestRegister.register, "/v1/items/sha-256:9432331d3343a7ceaaee46308069d01836460294c672223b236727a790acf786", WILDCARD);
-        assertThat(response.getStatus(), equalTo(307));
+        assertThat(response.getStatus(), equalTo(301));
```
gidsg: I think this change is not intended by the ADR, i.e. only the root paths should `301` other existing paths should continue to `307`. CC @arnau

arnau: agree

MatMoore: Have changed this back

PR #530, opened 2018-11-22
Upgrade gradle

gidsg:
### Context
Upgrade gradle (this is a prerequisite to eventually upgrading to Java 11)

* Upgrade to Gradle 4.10
* fork compass-gradle-plugin to publish version which supports Gradle 4.10 (see robfletcher/gradle-compass#77)
* Surpress Jruby warnings; I think these happen because we are no longer using an uber-jar in the fork, so these get pulled in. I don't think it's a security issue as it is just the CSS which gets built statically at compile time

### Guidance to review
`./gradlew` should work as before


gidsg:
closing as tests are failing.

PR #537, opened 2018-11-27
Fetch items by new hash or old hash

MatMoore:
### Context
Part of https://trello.com/c/JBenqGkg/3103-implement-rfc0010-item-hashing-algorithm-with-redactability

In order to support new objecthashes in v2 while continuing to maintain v1, a `Register` implementation needs to be able to query by both old and new hashes.

See the checklist on the card for the full list of steps to add item hashing.

- Add a new constructor to `Item` that allows it to store both hashes, and deprecate the old constructor
- When constructing an `Item` from a `JsonNode` (i.e. when consuming RSF) I set the new `blobHash` to the old style item hash for now. I will swap this out for the objecthash algorithm in a future PR.
- Rename `getItem` to `getItemByOldHash`, everywhere.
- Add back in new `getItem` methods which operate on the new hashes.
- Add `RegisterImpl` functional tests to test the lookups when an item has 2 hashes

### Guidance to review
I'm trying to split this into small steps because a lot of the codebase needs changing in order to support both v1 and v2 simultaneously. I think this is the most complicated part, so please review thoroughly.

Stuff I've deliberately not addressed here:
- All resource classes still use the old hash - I will raise a follow up PR to change this, once this one is merged
- `InMemoryItemDAO` isn't needed at all. I have a [separate PR](https://github.com/openregister/openregister-java/pull/535) to remove it.
- `BatchedPostgresDataAccessLayer` relies on hashes to cache item lookups. The main reason this exists is to assist with RSF loading, but I don't know if it's still needed now that we don't use indexes. For the time being I've not implemented any item caching based on the new hashes, but if we extend the new item hashing to `/load-rsf` I will benchmark both versions to see if we need this.

Naming things:
- For the new column/field I use "blob" throughout
- I still refer to the entity as `Item` everywhere except `BlobResource` and the DAOs, because the V1 API still uses this naming. When I first tried to rename all the internal code to "blob" it got messy, so I'd prefer to kick this can down the road until the point where v1 can be removed.


src/main/java/uk/gov/register/core/Item.java
```diff
@@ -17,25 +17,42 @@
...
     }
 
+    @Deprecated
+    public Item(HashValue oldHashValue, JsonNode content) {
+        // This should be removed
+        this(oldHashValue, oldHashValue, content);
+    }
+
+    public static HashValue objectHash(JsonNode content) {
+        // TODO
```
MatMoore: This will be implemented in a follow up PR

src/main/java/uk/gov/register/core/Item.java
```diff
@@ -17,25 +17,42 @@
...
-        this.hashValue = hashValue;
+    public Item(HashValue oldHashValue, HashValue blobHash, JsonNode content) {
+        this.oldHashValue = oldHashValue;
+        this.blobHash = blobHash;
         this.content = content;
     }
 
+    @Deprecated
+    public Item(HashValue oldHashValue, JsonNode content) {
+        // This should be removed
```
MatMoore: I can either remove this now (in a follow up PR), or leave it until we remove V1.

It's dangerous to leave it in, because it means we are creating items with the new hash set wrongly (even if its only in tests)

The reason I haven't changed it here in this PR is because it requires a lot of minor test changes, and it will clutter up the review

src/main/java/uk/gov/register/store/postgres/BatchedPostgresDataAccessLayer.java
```diff
@@ -168,11 +168,20 @@ public void addItem(Item item) {
...
         }
         
+        return isEntireRegisterBatched ? Optional.empty() : postgresDataAccessLayer.getItemByOldHash(hash);
+    }
+
+    @Override
+    public Optional<Item> getItem(HashValue hash) {
+        // TODO: this layer uses old hashes to store stuff in memory, so this lookup
+        // can't be cached. If this is caching is still necessary, migrate this to use
+        // new hashes.
```
MatMoore: I will look at this if/when we implement v2 of the `/load-rsf` endpoint. In the meantime this layer doesn't need to do anything because this method will only be called once per API request.

src/test/java/uk/gov/register/db/InMemoryItemDAO.java
```diff
@@ -32,6 +32,14 @@ public void insertInBatch(@BindItem Iterable<Item> items, String schema) {
         return Optional.ofNullable(items.get(new HashValue(SHA256, sha256Hash)));
     }
 
+    @Override
+    public Optional<Item> getItemByBlobHash(String blobHash, String schema) {
+        // TODO: this implementation assumes a single hash for all items, so it
+        // doesn't really work with the new interface. Consider whether we can scrap
+        // this class altogether.
```
MatMoore: See https://github.com/openregister/openregister-java/pull/535

src/test/java/uk/gov/register/functional/store/postgres/RegisterImplTransactionalFunctionalTest.java
```diff
@@ -72,6 +72,31 @@ public void tearDown() throws Exception {
         dbi.close(handle);
     }
 
+    @Test
+    public void getItemUsingNewHash() {
```
gidsg: Is it worth a `TODO` here to use the new hashing algorithm once implemented?

src/main/java/uk/gov/register/core/Item.java
```diff
@@ -17,25 +17,42 @@
...
-    public Item(HashValue hashValue, JsonNode content) {
-        this.hashValue = hashValue;
+    public Item(HashValue oldHashValue, HashValue blobHash, JsonNode content) {
+        this.oldHashValue = oldHashValue;
+        this.blobHash = blobHash;
         this.content = content;
     }
 
+    @Deprecated
+    public Item(HashValue oldHashValue, JsonNode content) {
```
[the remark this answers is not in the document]
MatMoore: Yeah that sounds better.

src/main/java/uk/gov/register/core/Item.java
gidsg: I think this is OK for now but can you add a `TODO` to the top the comment so it is flagged by the IDE.

src/test/java/uk/gov/register/functional/store/postgres/RegisterImplTransactionalFunctionalTest.java
MatMoore: I think this is ok actually... for the purposes of the test it doesn't matter what the hashes are, just that you can look up items by either one.

When I implement objecthash in `Item` I'll add separate tests for that.

src/main/java/uk/gov/register/core/Item.java
MatMoore: No problem, have added this

src/main/java/uk/gov/register/core/Item.java
```diff
@@ -17,25 +17,42 @@
     private static final CanonicalJsonMapper canonicalJsonMapper = new CanonicalJsonMapper();
 
     private final JsonNode content;
-    private final HashValue hashValue;
+    private final HashValue oldHashValue;
```
gidsg: could we rename this to `v1HashValue`?

MatMoore: ah missed one. Yeah.

PR #538, opened 2018-11-28
Include Java 11 removed libs

gidsg:
### Context
See https://github.com/openregister/openregister-java/pull/536 & https://github.com/openregister/incidents/pull/19

Include libraries removed in Java 11 to work on JRE 11
(for general context on this issue see https://stackoverflow.com/questions/48204141/replacements-for-deprecated-jpms-modules-with-java-ee-apis/48204154#48204154)

### Guidance to review
To test locally: in IntelliJ
import gradle changes
Edit RegisterRunner configuration
Set JRE to 11
App should still run and `/download-rsf` should work

I will also sanity check once this is on test environment that it works as expected on PaaS


gidsg:
@MatMoore erm, would depend exactly what the error is you are seeing but happy to look into this another time.

PR #541, opened 2018-11-30
Add a DropWizard command to populate v2 hash data

MatMoore:
### Context
Trello: https://trello.com/c/JBenqGkg/3103-implement-rfc0010-item-hashing-algorithm-with-redactability

Now that ORJ supports 2 concurrent hashing algorithms, we can populate the new hashes that will be read by the V2 API.

Add a command to populate `blob_hash` with objecthash hashes for all items, entries, and system entries. Previously `blob_hash` values were set to the V1 hashes.

### Guidance to review
- I made this a command rather than a migration, because I don't want it to block application startup if it takes a long time to run
- The command is run by passing `migrateHashes config.yaml` as arguments to the jar instead of `server config.yaml`
- It shouldn't matter too much if this command fails to execute, because only the `/next` API reads it, and the update is idempotent
- I used the imperative JDBI style instead of the declarative one, because I don't think the queries will be useful for anything else
- We can remove the command again once we're happy the V2 hashes are correct

### DIFF
--- a/src/main/java/uk/gov/register/commands/HashDataMigrationCommand.java
+++ b/src/main/java/uk/gov/register/commands/HashDataMigrationCommand.java
@@ -1,0 +1,65 @@
+package uk.gov.register.commands;
+
+import io.dropwizard.Application;
+import io.dropwizard.cli.EnvironmentCommand;
+import io.dropwizard.jdbi.DBIFactory;
+import io.dropwizard.setup.Environment;
+import net.sourceforge.argparse4j.inf.Namespace;
+import org.skife.jdbi.v2.DBI;
+import uk.gov.register.RegisterConfiguration;
+import uk.gov.register.configuration.ConfigManager;
+import uk.gov.register.configuration.DatabaseManager;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/commands/HashDataMigrationCommand.java
```diff
@@ -0,0 +1,64 @@
...
+import uk.gov.register.RegisterConfiguration;
+import uk.gov.register.configuration.ConfigManager;
+import uk.gov.register.configuration.DatabaseManager;
+import uk.gov.register.core.AllTheRegisters;
+import uk.gov.register.core.Register;
+import uk.gov.register.core.RegisterContext;
+import uk.gov.register.service.EnvironmentValidator;
+import uk.gov.register.util.JsonToBlobHash;
+
+public class HashDataMigrationCommand extends EnvironmentCommand<RegisterConfiguration> {
```
gidsg: Is it worth a `TODO:` to remove this once it's been run?

MatMoore:
Cool, I'll deploy this on Monday

PR #542, opened 2018-12-03
V2 Proofs and entry hashing

MatMoore:
### Context
This changes the data that goes into the merkle tree for the proof API to an objecthash of `key`, `timestamp`, `entry number` and `blob hash`, rather than a JSON string produced from the entry. This enables us to add additional metadata to entries without affecting the integrity of the merkle tree.

Trello: https://trello.com/c/4NkpexnZ/3102-implement-rfc0009-hashing-algorithm-for-entries

- [x] Rename V2 proof endpoints ([RFC 19](https://github.com/openregister/registers-rfcs/blob/master/content/proof-resource/index.md))
- [x] Add new MerkleLeafStore implementation that uses ObjectHash ([RFC 9](https://github.com/openregister/registers-rfcs/blob/master/content/entry-hash/index.md))
- [x] Make proof endpoints use the entry log directly instead of via a
  `ReadOnlyRegister` facade. Proofs are a very low level thing and I don't
think there is any reason for it to be in this higher level interface
- [x] Extract proof methods into a separate class and make `RegisterContext` build verifiable logs
- [x] Make V2 proof API use V2 proofs

### Guidance to review
The actual entry hashing part is very straightforward, but the example given in the [RFC](https://github.com/openregister/registers-rfcs/blob/master/content/entry-hash/index.md) is no longer correct due to [removing indexes](https://github.com/openregister/registers-rfcs/blob/master/content/indexes-removal/index.md), so I haven't compared the results to a known implementation. Instead I just stepped through the code in the debugger and made sure the component parts were correct.

The second commit is a very messy refactor of the proof code, and needs to be reviewed carefully. I made this change because the new entry hashes were very difficult to integrate without breaking proofs and RSF. I've tried to explain further in the commit message how I arrived at this solution.

I've removed a few tests but I think all of them are covered by others.

When reviewing this please pay special attention to the RSF code. Before merging we need to make sure that `/download-rsf` and `/load-rsf` still work the same and use V1 entry hashes.

We still need to add `/next/download-rsf` and `/next/load-rsf` in a future PR. The differences between the new and old RSF versions are:

- `assert-root-hash` uses V2 entry hashes
- `append-entry` identifies blobs using their V2 hashes

### DIFF
--- a/src/main/java/uk/gov/register/core/RegisterContext.java
+++ b/src/main/java/uk/gov/register/core/RegisterContext.java
@@ -1,6 +1,7 @@
 package uk.gov.register.core;
 
 import com.google.common.base.Throwables;
+import org.apache.commons.codec.digest.DigestUtils;
 import org.flywaydb.core.Flyway;
 import org.skife.jdbi.v2.DBI;
 import org.skife.jdbi.v2.Handle;
@@ -13,17 +14,21 @@ import uk.gov.register.db.RecordSet;
 import uk.gov.register.exceptions.FieldDefinitionException;
 import uk.gov.register.exceptions.NoSuchConfigException;
 import uk.gov.register.exceptions.RegisterDefinitionException;
+import uk.gov.register.proofs.EntryMerkleLeafStore;
+import uk.gov.register.proofs.V1EntryMerkleLeafStore;
 import uk.gov.register.service.EnvironmentValidator;
 import uk.gov.register.service.ItemValidator;
 import uk.gov.register.store.DataAccessLayer;
 import uk.gov.register.store.postgres.BatchedPostgresDataAccessLayer;
 import uk.gov.register.store.postgres.PostgresDataAccessLayer;
+import uk.gov.verifiablelog.VerifiableLog;
 import uk.gov.verifiablelog.store.memoization.InMemoryPowOfTwoNoLeaves;
 import uk.gov.verifiablelog.store.memoization.MemoizationStore;
 
 import java.io.IOException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 
 public class RegisterContext implements
@@ -76,14 +81,14 @@ public class RegisterContext implements
     public EntryLog buildEntryLog() {
         DataAccessLayer dataAccessLayer = getOnDemandDataAccessLayer();
 
-        return new EntryLogImpl(dataAccessLayer, memoizationStore.get());
+        return new EntryLogImpl(dataAccessLayer);
     }
 
     public Register buildOnDemandRegister() {
         DataAccessLayer dataAccessLayer = getOnDemandDataAccessLayer();
 
         return new RegisterImpl(registerId,
-                new EntryLogImpl(dataAccessLayer, memoizationStore.get()),
+                new EntryLogImpl(dataAccessLayer),
                 new ItemStoreImpl(dataAccessLayer),
                 new RecordSet(dataAccessLayer),
                 itemValidator,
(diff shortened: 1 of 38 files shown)


src/main/java/uk/gov/register/db/Factories.java
```diff
@@ -20,6 +21,20 @@ public void dispose(T instance) {
         }
     }
 
+    public static class EntryLogFactory extends SimpleFactory<EntryLog> {
```
MatMoore: This might not be needed any more - will double check tomorrow

src/main/java/uk/gov/register/db/EntryMerkleLeafStore.java
```diff
@@ -1,42 +1,41 @@
...
     private final EntryIterator entryIterator;
 
-
     public EntryMerkleLeafStore(EntryIterator entryIterator) {
         this.entryIterator = entryIterator;
     }
-
     @Override
     public byte[] getLeafValue(int i) {
         return bytesFromEntry(entryIterator.findByEntryNumber(i + 1));
```
gidsg: what's the purpose of the `+ 1` here, is it something to do with things being 0 indexed?

src/main/java/uk/gov/register/serialization/handlers/AddItemCommandHandler.java
```diff
@@ -17,7 +18,7 @@ public AddItemCommandHandler() {
     }
 
     @Override
-    protected void executeCommand(RegisterCommand command, Register register) {
+    protected void executeCommand(RegisterCommand command, Register register, ProofGenerator _proofGenerator) {
```
gidsg: does not seem this `_proofGenerator` argument is used?

src/main/java/uk/gov/register/db/EntryMerkleLeafStore.java
MatMoore: Entry number starts at 1, but MerkleLeafStore implementations should be zero indexed.

src/main/java/uk/gov/register/resources/v2/VerifiableLogResource.java
```diff
@@ -65,4 +72,6 @@ private void validateConsistencyProofParams(int totalEntries1, int totalEntries2
             throw new BadRequestException("Invalid parameters");
         }
     }
+
```
gidsg: very minor but we can remove this whitespace.

src/main/java/uk/gov/register/serialization/handlers/AddItemCommandHandler.java
gidsg: I guess we are trying to indicate an unused argument here with the underscore? In Java would just stick with `camelCase` even for an unused override variable like this.

src/main/java/uk/gov/register/core/RegisterContext.java
```diff
@@ -139,6 +144,26 @@ private DataAccessLayer getOnDemandDataAccessLayer() {
...
+            VerifiableLog verifiableLog = new VerifiableLog(
+                    DigestUtils.getSha256Digest(),
+                    new V1EntryMerkleLeafStore(entryIterator),
+                    memoizationStore.get());
+            return callback.apply(verifiableLog);
+        });
+    }
+
+    public <R> R withVerifiableLog(Function<VerifiableLog, R> callback) {
+        return getOnDemandDataAccessLayer().withEntryIterator(entryIterator -> {
```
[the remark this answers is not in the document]
MatMoore: Can you post a snippet of what you have in mind?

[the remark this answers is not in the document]
gidsg: sorry, don't worry about it. I had a go but couldn't get what I was thinking about to compile!

PR #547, opened 2018-12-05
Implement RFC0025 Snapshot resource for records

gidsg:
### Context
https://github.com/openregister/registers-rfcs/blob/master/content/snapshot-resource/index.md

Implements snapshot resource for records. Note this does not implement the extra log size features.

### Guidance to review
Note, this adds `_id`  at the end of the JSON array and at the start of the CSV as it was easiest to serialise that way. I believe this is spec compliant as per https://spec.openregister.org/v2/rest-api/records but let me know if you feel strongly about this

# TODO
- [x] update conformance tests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 csvvalidator==1.2
 jsonschema==2.5.1
--e git+https://github.com/openregister/conformance-test.git@26675dac36c7786189623dc754df7de0356210fb#egg=openregister_conformance-master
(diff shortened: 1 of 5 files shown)


arnau:
> Note, this adds _id at the end of the JSON array and at the start of the CSV as it was easiest to serialise that way. I believe this is spec compliant as per https://spec.openregister.org/v2/rest-api/records but let me know if you feel strongly about this

It is perfectly fine as per the spec. And I'm fine with it as well :)

PR #551, opened 2018-12-07
Change the structure of the v2 /blobs endpoint to an array.

MatMoore:
### Context
Trello: https://trello.com/c/HTgdMWyu/3149-finish-blobs-endpoint

Until now this endpoint has been limited to the first 100 entries. The response structure was also inconsistent with other collection resources.

This PR diverges from [ADR 8: Items resource](https://github.com/openregister/openregister-java/blob/master/doc/arch/adr-008-items-resource.md) so I've added [another ADR](https://github.com/openregister/openregister-java/blob/d4f4da1a7bba41d3f6c98a3236d1593d42ffd8ae/doc/arch/adr-012-prefer-arrays-for-collections.md) to summarise the discussion I had with @arnau about this.

Change the response structure, use `_id` for the blob hash, and add pagination.

The migration tries to put existing items in a logical order, by looking at the first entry associated with each item and then ordering each item by a combination of entry number and type (system or user). Even though the blobs belonging to system entries are not visible in the response, they still get numbered. I order these ahead of the user entries to match `/download-rsf`.

`item_order` can have gaps in it, because the presence of system entries effectively creates gaps in the visible blobs, and because the field is a serial field in the database. If inserts get rolled back, then postgres still increments the sequence.

This is deliberately different from the way we implemented pagination for entries and records, but is still consistent with the spec:

- The ordering remains stable when you add more blobs to the database
- Consumers should use `Link` headers as is, rather than construct their own
- Implementations don't need to provide a `rel=previous` link

Before merging:
- [x] Add ADR
- [x] Point the conformance tests to a [new version](https://github.com/openregister/conformance-test/pull/35)

### Guidance to review
- The migration should work for any register
- `/blobs` should return a link to the next page
- The blobs resource should order older items first, so that pages don't change when new stuff is added


arnau:
I just opened this PR https://github.com/openregister/registers-rfcs/pull/44 to address the lack of RFC for the blob resource change. When approved I'll change the specification accordingly.

doc/arch/adr-012-prefer-arrays-for-collections.md
```diff
@@ -0,0 +1,40 @@
...
+## Decision
+
+1. The JSON representation for any collection resource should be an array of objects, unless there is a good reason to
+   break this rule
+2. The special value `_id` should be used to include identifiers in the objects
+
+[RFC 25](https://github.com/openregister/registers-rfcs/blob/master/content/snapshot-resource/index.md) established this
+pattern for records specifically.
+
+We intend to make the blobs resource consistent with this decision, although the specification doesn't reflect this yet.
```
arnau: The spec does reflect this now

src/main/java/uk/gov/register/store/postgres/PostgresDataAccessLayer.java
```diff
@@ -87,7 +87,7 @@ public int getTotalEntries(EntryType entryType) {
             default: throw new RuntimeException("Entry type not recognised");
         }
     }
-
+
```
gidsg: very minor, but this looks like a whitespace only change that could be removed.

src/main/java/uk/gov/register/views/v2/BlobView.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+public class BlobView implements CsvRepresentationView<Map<String, FieldValue>> {
+    private final Iterable<Field> fields;
+    private final Map<String, FieldValue> fieldValueMap;
+
+    public BlobView(Item item, final Map<String, Field> fieldsByName, final ItemConverter itemConverter) {
+        this.fields = fieldsByName.values();
+        this.fieldValueMap = new HashMap(itemConverter.convertItem(item, fieldsByName));
```
gidsg: If you change this to `new HashMap<>` it fixes an IntelliJ warning about unchecked assignment.

src/main/java/uk/gov/register/resources/v2/BlobResource.java
```diff
@@ -58,11 +69,21 @@ public ItemView getBlobDataByHex(@PathParam("blob-hash") String blobHash) throws
...
+        int limit =  optionalLimit.map(IntParam::get).orElse(100);
 
-        // TODO: allow this resource to be paginated
-        return buildBlobListView(items.stream().limit(100).collect(Collectors.toList()));
+        Collection<Item> items = register.getUserItemsPaginated(start, limit + 1);
+
+        boolean hasNext = (items.size() == limit + 1);
+
+        if(hasNext) {
+            Item finalBlob = items.stream().skip(limit).findFirst().get();
```
gidsg: it is dangerous to call `.get()` on an `Optional` can we do something similar to https://github.com/openregister/openregister-java/blob/292138e102e760434a987fcc1cef851ba98110e8/src/main/java/uk/gov/register/resources/v1/ItemResource.java#L45 ?

src/main/java/uk/gov/register/resources/v2/BlobResource.java
```diff
@@ -74,11 +95,16 @@ public BlobListView listBlobs() throws FieldConversionException {
         return register.getFieldsByName();
     }
 
-    protected ItemView buildItemView(Item item) {
-        return new ItemView(item, register.getFieldsByName(), this.itemConverter);
+    protected BlobView buildBlobView(Item item) {
```
gidsg: can we make this `private`?

src/main/java/uk/gov/register/core/Item.java
```diff
@@ -20,6 +20,7 @@
     private final JsonNode content;
     private final HashValue v1HashValue;
     private final HashValue blobHash;
+    private final Optional<Integer> itemOrder;
```
gidsg: it does seem a bit weird that this is a property of the Item itself rather than the view where it is paginated. But happy to go with this for now and re-visit if we get time.

src/main/java/uk/gov/register/views/v2/BlobView.java
gidsg: I wonder if we should use a `LinkedHashMap` here to preserve the order, else I *think* we risk an inconsistently ordered response? Not that we make any guarantees about this but it is good to be consistent for caching etc.

src/main/java/uk/gov/register/resources/v2/BlobResource.java
MatMoore: We could but this should never get called if you're falling into this block because it checks the size first. I'm just trying to access the last item in the list here but I couldn't figure out a nice way to do it. Is there a cleaner way to rewrite this whole block?

PR #560, opened 2018-12-19
Implement RFC022 (Facets) [WIP]

gidsg:
### Context
https://github.com/openregister/registers-rfcs/blob/master/content/facets-endpoint/index.md
https://trello.com/c/z5lEkxuN/3155-implement-rfc0022-facets-endpoint

Add Facets implementation using new URL structure

#### TODO

Note, this inherits the V1 bug/limitation of having a fixed size and not supporting pagination on this endpoint

- [ ] support pagination

### DIFF
--- a/src/main/java/uk/gov/register/resources/v2/RecordResource.java
+++ b/src/main/java/uk/gov/register/resources/v2/RecordResource.java
@@ -74,19 +74,7 @@ public class RecordResource {
         }
         return new EntryListView(allEntries);
     }
-
-    @GET
-    @Path("/{key}/{value}")
-    @Produces({
-            MediaType.APPLICATION_JSON,
-            ExtraMediaType.TEXT_CSV,
-    })
-    @Timed
-    public RecordListView facetedRecords(@PathParam("key") String key, @PathParam("value") String value) throws FieldConversionException {
-        List<Record> records = register.max100RecordsFacetedByKeyValue(key, value);
-        return new RecordListView(records, register.getFieldsByName());
-    }
-
+    
     @GET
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/register/resources/v2/RecordResource.java
```diff
@@ -113,8 +101,16 @@ private IndexSizePagination setUpPagination(@QueryParam(IndexSizePagination.INDE
...
-    private RecordListView getRecordsView(int limit, int offset) throws FieldConversionException {
-        List<Record> records = register.getRecords(EntryType.user, limit, offset);
+    private RecordListView getRecordsView(int limit, int offset, Optional<String> attributeName, Optional<String> attributeValue) throws FieldConversionException {
+        List<Record> records;
+
+        if(attributeName.isPresent() && attributeValue.isPresent()) {
+
+            records = register.getRecordsFacetedByKeyValue(attributeName.get(), attributeValue.get(), limit, offset);
+        }
+        else {
```
MatMoore: Should we return an error if they pass in only one of the two params?

gidsg:
@MatMoore since this is an optional feature, I don't think we should include it in conformance tests.

MatMoore:
@gidsg Ah I forgot it was optional, yeah let's leave out then.

gidsg:
I have reverted the half finished pagination changes as it was proving non-trivial to get the total count. If it's OK I'd like to merge this as-is mirroring the V1 behaviour and then make a spin off card to address pagination properly. How does that sound to you @MatMoore?

PR #561, opened 2018-12-19
Add V2 Context endpoint

MatMoore:
### Context
This is the new context endpoint described in [RFC 18](https://github.com/openregister/registers-rfcs/blob/master/content/context-resource/index.md)

Implement the endpoint.

Note that I have made `start-date` optional (contradicting the spec) because it's possible with ORJ to create an empty register that has not actually started yet. The field should be populated for all registers we've released to beta.


src/main/java/uk/gov/register/views/v2/ContextView.java
```diff
@@ -0,0 +1,134 @@
...
+    public String getRootHash() {
+        return rootHash;
+    }
+
+    @JsonProperty
+    public Map<String, Integer> getStatistics() {
+        Map<String, Integer> result = new LinkedHashMap<>();
+        result.put("total-entries", this.totalEntries);
+        result.put("total-records", this.totalRecords);
+        result.put("total-items", -1);
```
gidsg: Is this value a placeholder?

MatMoore: Yeah, it was. Added in a query for this now.

src/main/java/uk/gov/register/core/Register.java
```diff
@@ -6,5 +6,6 @@
 @Contract
 public interface Register extends RegisterReadOnly {
     void addItem(Item item);
+
```
gidsg: looks like this file can be reverted as there is no code change.

PR #564, opened 2018-12-31
Upgrade to Gradle 4.10.3

gidsg:
### Context
Previously we were running a no longer supported version of Gradle (2.6)

Upgrade to Gradle 4.10.3

### Guidance to review
Build should pass in Travis CI and deploy successfully in Codepipeline

(Note: this fixes any deprecation warnings for Gradle 5.0 which supports JDK 11, however, due to issues when upgrading Dependency Check plugin this PR sticks to version 4.10)


src/test/java/uk/gov/register/functional/db/RecordQueryFunctionalTest.java
```diff
@@ -91,6 +94,7 @@ public void setup() throws IOException {
     @After
     public void tearDown() {
         dbi.close(handle);
+        register.wipe();
```
gidsg: This test was previously not cleaning database state which caused subsequent tests to fail depending in what order tests were run.

PR #571, opened 2019-01-04
fix guava throwables deprecation

gidsg:
### Context
future proof app

Fix Guava deprecation warning, see https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate#why-we-deprecated-throwablespropagate for full details

### Guidance to review
Build should pass in Travis CI app should still run

### DIFF
--- a/src/main/java/uk/gov/register/RegisterApplication.java
+++ b/src/main/java/uk/gov/register/RegisterApplication.java
@@ -1,7 +1,6 @@
 package uk.gov.register;
 
 import com.codahale.metrics.health.HealthCheckRegistry;
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import io.dropwizard.Application;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/register/RegisterApplication.java
```diff
@@ -20,8 +19,19 @@
 import uk.gov.ida.dropwizard.logstash.LogstashBundle;
 import uk.gov.register.auth.BasicAuthFilter;
 import uk.gov.register.auth.RegisterAuthDynamicFeature;
-import uk.gov.register.configuration.*;
-import uk.gov.register.core.*;
+import uk.gov.register.configuration.ConfigManager;
```
gidsg: these verbose imports are just due to following GDS way policy when updating files: https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

PR #580, opened 2019-02-04
upgrade dropwizard to 1.3.8

gidsg:
### Context
master build was failing due to dependency check warning

Upgrade to dropwizard 1.3.8 (and associated changes to support this)
Note: I have removed `dropwizard-graphite` as I don't believe this is being used (since the move to PaaS)

### Guidance to review
Build should pass on travis

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,6 @@ dependencies {
     compile(postgresClient) {
         exclude group: 'org.slf4j'
     }
-    compile "com.google.guava:guava:27.0-jre"
     compile jetty
 
     testCompile junit, mockito, dropwizardTest, wiremock
--- a/build.gradle
+++ b/build.gradle
@@ -67,9 +67,6 @@ dependencies {
     compile(postgresClient) {
         exclude group: 'org.slf4j'
     }
-    compile ("io.dropwizard:dropwizard-lifecycle"){
-        exclude group: 'com.google.guava', module: 'guava'
-    }
     compile "com.google.guava:guava:27.0-jre"
     compile jetty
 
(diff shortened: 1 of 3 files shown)


external-dependencies.gradle
```diff
@@ -1,12 +1,12 @@
 ext {
-    dropwizard_version = '1.1.0'
-    dropwizard_metrics_version = '3.1.0'
+    dropwizard_version = '1.3.8'
```
danworth: 👍

src/main/java/uk/gov/register/thymeleaf/ThymeleafViewRenderer.java
```diff
@@ -68,7 +68,7 @@ public void configure(Map<String, String> options) {
     }
 
     @Override
-    public String getSuffix() {
-        return suffix;
+    public String getConfigurationKey() {
```
danworth: Could you help me understand why this is necessary please?

build.gradle
```diff
@@ -67,10 +67,6 @@ dependencies {
     compile(postgresClient) {
         exclude group: 'org.slf4j'
     }
-    compile ("io.dropwizard:dropwizard-lifecycle"){
-        exclude group: 'com.google.guava', module: 'guava'
```
danworth: I'm unfamiliar with gradle, does removing `compile "com.google.guava:guava:27.0-jre"` mean that it will no longer be included in the final artefact/jar? Could you please help me understand?

gidsg: Guava is already included in the jar, this was just overriding the old version previously included by dropwizard i.e. this is the pattern described in https://stackoverflow.com/a/30729087/238230

src/main/java/uk/gov/register/thymeleaf/ThymeleafViewRenderer.java
danworth: 👍 thanks!

build.gradle
danworth: ah ok that makes sense, I can see how its working now :)

